PR #210, opened 2014-10-31
ADD: Ability to send an optional UncaughtExceptionHandler to the context

erichfosse:
I've added a couple of constructors, and I'm testing for null values where necessary, so one can choose whether or not to use the context with an UncaughtExceptionHandler.
As previously mentioned, I'm not sure if this is best practice, but it makes it easy to close and re-open socket connections when the thread terminates. Hope it's of some use to you.


trevorbernard:
Tried rebuilding and it fails the style guide.

PR #230, opened 2014-11-24
Resource deallocation

fredoboulo:
Here is a sample test as well as the revert of the commit for the closing of the sockets.
Please note that the unit test does not end with Test so it will not be running in automatic testing, as this kind of stress test may need to run for a long period of time.

This exact same test can be used on my machine to detect resource deallocation problems in sockets as well as in polling

### DIFF
--- a/src/main/java/zmq/Ctx.java
+++ b/src/main/java/zmq/Ctx.java
@@ -149,6 +149,8 @@ public class Ctx
         termMailbox.close();
 
         tag = 0xdeadbeef;
+
+        zmq.ZMQ.closePollSelector();
     }
 
     //  Returns false if object is not a context.
--- a/src/main/java/zmq/Ctx.java
+++ b/src/main/java/zmq/Ctx.java
@@ -344,6 +344,7 @@ public class Ctx
         try {
             int tid = socket.getTid();
             emptySlots.add(tid);
+            slots[tid].close();
             slots[tid] = null;
 
             //  Remove the socket from the list of sockets.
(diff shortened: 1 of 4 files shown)


src/main/java/zmq/Ctx.java
```diff
@@ -149,6 +149,8 @@ private void destroy() throws IOException
         termMailbox.close();
 
         tag = 0xdeadbeef;
+
+        zmq.ZMQ.closePollSelector();
```
daveyarwood: I'm almost 2 years late to this conversation, but is there a reason that the PollSelector needs to be ThreadLocal? One thing I've read about ThreadLocal is that it makes it easy to introduce a memory leak. I think we might be seeing an example of this with the "Too many files open" problems.

src/main/java/zmq/Ctx.java
```diff
@@ -344,6 +346,7 @@ public void destroySocket(SocketBase socket)
         try {
             int tid = socket.getTid();
             emptySlots.add(tid);
+            slots[tid].close();
```
daveyarwood: Seems like we don't want to revert this fix. I would advocate for removing this commit.

src/main/java/zmq/Ctx.java
fredoboulo: I'm also late...
I would tend to give it an historical reason, @miniway could tell us more about that.
The tests I made on this library advocate against the use of ThreadLocal and finalize() to release the selector resource. This was also one big motivation to create the ZProxy class: having a better control on the proxy operations and suppress this memory leak for my own uses.

daveyarwood:
Superceded by #386.

PR #239, opened 2015-02-05
Change ZMQ.bind() method to return void.

yebenes:
ZMQ.bind method changed because is used when the port is known and  no sense return the port. Also, with this change, jeromq would have the same contract that JZMQ and ZeroMQ-Scala-Binding. Finally, most importantly, this change would use scala-zeromq library with jeromq, for now, this library only allows use with the two mentioned above libraries.

[Review on Reviewable](https://reviewable.io/reviews/zeromq/jeromq/239)


sjohnr:
+1. Is there a reason we can't merge this?

yebenes:
Thanks! :)

PR #242, opened 2015-02-27
ZFrame.toString() – Break loop on finding the first non-printable character

cPu1:
Whether a byte stream is printable or not can be determined on finding the first ASCII value that is not printable. Without `break`, a full iteration over the byte array will be performed.


miniway:
nice catch!

Issue #172, opened 2014-06-27
Spinning in Reaper Thread


cwolfinger:
Hi - The code seems like it can spin in the reaper thread:

"reaper-1" prio=10 tid=0x00002aaab8c1d800 nid=0x18f3 runnable [0x00000000497e1000]
java.lang.Thread.State: RUNNABLE
at sun.nio.ch.NativeThread.current(Native Method)
at sun.nio.ch.SourceChannelImpl.read(SourceChannelImpl.java:166)
- locked <0x00000000c3a0c3f8> (a java.lang.Object)
  at zmq.Signaler.recv(Signaler.java:160)
  at zmq.Mailbox.recv(Mailbox.java:107)
  at zmq.SocketBase.process_commands(SocketBase.java:799)
  at zmq.SocketBase.in_event(SocketBase.java:907)
  at zmq.Poller.run(Poller.java:231)
  at java.lang.Thread.run(Thread.java:722)

We have found it goto 100% CPU.  Seems like the channel is non blocking and it runs in a tight loop and therefore spins.

Has anyone else seen this?

simongregorebner:
I am not sure whether this is related to the problem discussed earlier, however I am able to produce 100% CPU load by closing a ZMQ socket in a thread that has the interrupted state set.
The code to reproduce this is as follows:

```
Context context = ZMQ.context(1);
Socket socket = context.socket(ZMQ.PUSH);
socket.bind("tcp://*:8888");
		
// Set interrupt state	
Thread.currentThread().interrupt();
// This will generate 100% CPU load

socket.close(); // There will be an exception if interrupted is set
context.close();
```

So far I had no time to look into the exact reason for this, however this might help you ...

daveyarwood:
I think #248 solved this. Feel free to reopen if that is not the case.

PR #248, opened 2015-05-13
Fix Spinning in Reaper Thread

miniway:
To address Issue #172, signaler clear interrupted thread state right before the pipe write. Even though this, `Pipe` could still be `ClosedByInterruptException` at write and very hard to handle it.


trevorbernard:
:+1: :shipit:

PR #263, opened 2015-07-01
Poller now keeps index when un-registering sockets

damienlevin:
Hi,

I was experiencing some issues using the library and found out that the indexing done by the Poller is not preserved when un-registering sockets. jzmq doesn't seems to have this behavior.

The following commit seems to fix my issues. Let me know what you think.

Best

Damien


daveyarwood:
I think #374 may fix this issue.

PR #269, opened 2015-07-16
Change to allow org.zeromq to be OSGi bundle and work in OSGi server

gerring:
This change allows a Java developer to import the project as an OSGi bundle called 'org.zeromq'. It might be unfriendly to non-eclipse platforms because of relaxing the gitignore but would allow the plugin to import as a bundle - but please consider it! :-)


trevorbernard:
I'd prefer a solution that involved generating the bundle in maven so it works for everybody, repeatably. Are you interested in porting your current solution to the maven bundle plugin?

gerring:
Hi Trevor,

Yes interested but don't know what to do, would need advice. I would have
thought this change agnostic to maven i.e. once META-INF is added, it
should be an OSGi bundle for everyone already.

Matt

daveyarwood:
Closing this PR, as it's getting stale. @gerring / @trevorbernard: feel free to reopen if you'd like to continue the conversation above.

PR #274, opened 2015-08-03
Slight refactorings.

sbanacho:
These are some minor changes I made while I was browsing the code to understand it.  They should have no impact other than making the code clearer (hopefully).

ZContext: make the parameters set in the constructor final, and remove setters for those parameters.
Ctx:  move the code block for initialization of slots to a separate function, to reduce the size of createSocket().
Reaper:  move duplicated code block to a function.  Also, increment reap count before starting the reap instead of after.

### DIFF
--- a/src/main/java/org/zeromq/ZContext.java
+++ b/src/main/java/org/zeromq/ZContext.java
@@ -42,17 +42,17 @@ public class ZContext implements Closeable
     /**
      * Reference to underlying Context object
      */
-    private volatile Context context;
+    private volatile Context context;  //  Created lazily, use getContext() to access.
 
     /**
      * List of sockets managed by this ZContext
      */
-    private List<Socket> sockets;
+    private final List<Socket> sockets;
 
     /**
      * Number of io threads allocated to this context, default 1
      */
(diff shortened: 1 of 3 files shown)


daveyarwood:
I like the spirit of this PR, and the code looks good to me. 👍 

I also agree with @trevorbernard about deprecating public functions rather than removing them.

I'd like to get this PR moving, but I can't answer this question with 100% authority:

> In this case do you prefer I add the deprecated functions back as no-ops, or should I also revert the behavior that the fields are modifiable?

If it were up to me, I would say:

- Add the deprecated functions back as no-ops (perhaps print a deprecation warning).
- Do not revert the behavior that the fields are modifiable. This will possibly break someone's code, but holding true to the spirit of this PR, any code that modifies these fields (e.g. creating a shadow context and setting "main" to true) is bad code and deserves to be rewritten.

@trevorbernard Would you agree?

daveyarwood:
Superceded by #400.

sbanacho:
Thanks for following up on this.

daveyarwood:
Happy to help!

PR #298, opened 2015-11-02
Problem: Example does not compile since ZHelper class does not exists.

chilicat:


miniway:
hmm, `ZHeler.java` exists in the same directory.

chilicat:
Oh I see. I started to copy the example from the zmq guide (http://zguide.zeromq.org/java:lbbroker). Generally it would be better if examples have no shared code. But now I know. Thank you.

Issue #307, opened 2015-12-22
Make extra byte copy obsolete


markif:
When I implemented #305, I realized that Msg.data() creates an extra copy when users use DirectByteBuffers or HeapByteBuffers with position>0 or limit!=capacity. It would be possible to save this extra copy when EncoderBase and DecoderBase would use ByteBuffer instead of byte[]. 
The aim of this issue it to 
- provide code that makes this extra copy obsolete
- set the stage to make #281 possible

PR #308, opened 2015-12-22
#307

markif:
Pull request for #307.
Aim is to avoid extra copy when Msg contains a DirectByteBuffers or HeapByteBuffers with position>0 or limit!=capacity. To make this possible, EncoderBase and DecoderBase also use ByteBuffers and ByteBuffer in Msg needs to keep track of its initial positions (the problem was that it allows relative operations on its ByteBuffer but removing these methods would break compatibility with existing code therefore I decided to use write-/readIndex).

### DIFF
--- a/src/main/java/org/zeromq/ZMQ.java
+++ b/src/main/java/org/zeromq/ZMQ.java
@@ -1404,7 +1404,7 @@ public class ZMQ
             zmq.Msg msg = base.recv(flags);
 
             if (msg != null) {
-                buffer.put(msg.data());
+                buffer.put(msg.buf());
                 return msg.size();
             }
 
--- a/src/main/java/zmq/DecoderBase.java
+++ b/src/main/java/zmq/DecoderBase.java
@@ -35,11 +35,7 @@ import java.nio.ByteBuffer;
 public abstract class DecoderBase implements IDecoder
 {
     //  Where to store the read data.
-    private byte[] readBuf;
-    private int readPos;
-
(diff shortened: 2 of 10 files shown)


miniway:
Thanks for the contribution. Did you see any performance gain on the change? Could you please compare the performance different with simple bt running the some tests under the `perf`

markif:
Not yet. But I am happy to do so...

BTW: I expect performance gains for situations where people use DirectByteBuffers and/or want to send only parts of a bigger byte[]/ByteBuffer (in these cases an extra copy step will not be necessary anymore). For other use cases, the hope is to have similar performance.

markif:
The attachement shows the performance comparison beween the original implementation (before my commit) and the implementation using ByteBuffers.

General insights :
- ByteBuffers are less performant for small message sizes (between 0.89 and 0.96 of original performance).
- ByteBuffers have same performance for bigger message sizes (>=10K)
- For special use cases (e.g., DirectByteBuffers or HeapBuffers with pos != 0) the new implementation has better performance (up to 2.4 times the original speed, which is not surprising as it is somewhat tailored to our use case where we use DirectByteBuffers with pos != 0 and Msg sizes up to 16MB).

Based on this insight I am uncertain about the next steps as the changes show there advantage only for big message sizes (which might not be the main use case for jeromq). 

None the less, I did some performace tweeks which I address in #310

[Comparison.pdf](https://github.com/zeromq/jeromq/files/70975/Comparison.pdf)

Issue #281, opened 2015-08-20
Possibility to Define ByteBuffer Type for Msg


markif:
We use your great library to send big data blobs (one message around 16MB). After receiving the data, we need to buffer the messages for a short period. To overcome OutOfMemoryErrors we use DirectByteBuffers.

Have you considered providing users the possibility to define which type of ByteBuffer Msg should use when receiving data (e.g. users can specify a IntFunction<ByteBuffer> (or a similar interface for < Java 8) that returns a ByteBuffer based on the requested capacity)? In our case, this feature would prevent an additional copy step (HeapByteBuffer of Msg -> DirectByteBuffer).

miniway:
Thank you for the idea. Even though we're trying our best to make JeroMQ compatible and interchangeable with JZMQ, there's one non compatible feature at the` StreamEngine` and the related configuration. https://github.com/zeromq/jeromq/blob/master/src/main/java/zmq/Options.java#L301

Please send us a pull request on your idea, we could start deeper discussion on it.

PR #309, opened 2015-12-22
MsgAllocator allows users to define how Msg are created

markif:
MsgAllocator allows users to define how Msg are created (respectively Msg's ByteBuffer). This allows for more flexibility (e.g., this allows us to directly use DirectByteBuffers and initialize them with some extra space at their start where header information will be stored in a later step – saving us some extra copying work). 
We should probably first discuss #308 as this pull request builds on it.

### DIFF
--- a/src/main/java/zmq/Decoder.java
+++ b/src/main/java/zmq/Decoder.java
@@ -108,7 +108,7 @@ public class Decoder extends DecoderBase
 
             }
             else {
-                inProgress = new Msg(size - 1);
+                inProgress = getMsgAllocator().allocate(size - 1);
             }
 
             nextStep(tmpbuf, 1, FLAGS_READY);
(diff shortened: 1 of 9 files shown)


markif:
Hey Constantin

I realized that there was a problem with TestZActor and TestReqRouterThreadedTcp (both failing once out of four runs both times with different jre). Might be due to a Threading issue. How should I proceed with this (as you decided to merge the pull requests)?

c-rack:
Yes, saw this. Both seem to be threading issues for me as well and I don't think they are related to your changes. Travis is sometimes overloaded. We see this in other projects as well. I will restart the build later today and see if tests passes. If you want to look into it to see why they fail, you are welcome.

Regarding your changes: you have clearly described a problem (#281) and provided patches (to prepare) to solve that problem step-by-step. That's why I have merged your pull requests.

Only thing that could be improved IMHO is the naming of commit messages and pull requests. Try to describe the problem your commit / PR solves rather than describing the change itself. Example: instead of "MsgAllocator allows users to define how Msg are created" you could write "Problem: Msg allocation is not configurable" or similar. Same applies for your pull requests #308 and #306 :-)

A last hint regarding pull requests: if you include "fixes #issue" in either your commit or pull request message, that issue will be automatically closed when pull request is merged:
https://help.github.com/articles/closing-issues-via-commit-messages/

Otherwise, you'd have to close your issues (like #307) by yourself :-)

Thanks for your contributions and feel free to proceed! :+1:

markif:
Ok. Thanks for your advice.

PR #317, opened 2016-01-16
Fix travis ci build

trevorbernard:


c-rack:
Just had the same idea ;-)

Issue #327, opened 2016-02-15
Problem: JeroMQ still uses LGPL license


c-rack:
Solution: migrate to MPLv2, as explained here http://zeromq.org/area:licensing

* [x] @trevorbernard
* [x] @miniway
* [x] @markif
* [x] @fernandezpablo85
* [x] @sjohnr
* [x] @davipt
* [x] @ealgell
* [x] @HeinrichHartmann
* [x] @capacman
* [x] @crocket
* [x] @ipechorin
* [x] @nicholassm
* [x] @msteinhoff
* [x] @markrileybot
* [x] @sbanacho
* [x] @kevinkreiser
* [x] @ifesdjeen
* [x] @hintjens
* [x] @vsg
* [x] @cbusbey
* [ ] @imleon
* [x] @kcarlson
* [ ] @kedzie
* [ ] @cossin
* [x] @devinrsmith
* [x] @pavelbucek
* [x] @yebenes
* [x] @zh217
* [x] @arashthearcher
* [x] @gsnewmark
* [x] @baboune

Could you kindly reply
> I hereby agree to publish my contributions to JeroMQ under the MPLv2 license

?

If you have any objection to relicensing your contributions, let us know the reasons. Thanks!

benjyz:
@c-rack my contribution was a trivial commit and negligible, just remove my name. I don't want to be associated with this project in any form. please don't contact me again.

hintjens:
I hereby agree to publish my contributions to JeroMQ under the MPLv2 license.

Thanks for doing this.

davipt:
I hereby agree to publish my contributions to JeroMQ under the MPLv2 license. 

Thanks for everything. 


--
Bruno D. Rodrigues

No dia 15/02/2016, às 22:42, Constantin Rack <notifications@github.com> escreveu:

> I hereby agree to publish my contributions to JeroMQ under the MPLv2 license

kevinkreiser:
I hereby agree to publish my contributions to JeroMQ under the MPLv2 license.

capacman:
I hereby agree to publish my contributions to JeroMQ under the MPLv2 license

On Tue, Feb 16, 2016 at 12:59 AM, Kevin Kreiser <notifications@github.com>
wrote:

> I hereby agree to publish my contributions to JeroMQ under the MPLv2
> license.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/zeromq/jeromq/issues/327#issuecomment-184432270>.
>



[...]

PR #354, opened 2016-08-18
Revert commits for #327

parente:
For #327


parente:
What next? Does someone else add the same fixes back in a different way?

The test suite seems to pass still, but unreliably.

trevorbernard:
Code speaks, but I would argue what problems your changes solve?

PR #358, opened 2016-09-08
Change license from LGPL to MPLv2

### LINKED ISSUE #326: What is the status of the MPL license changes to this project?
The MPL statement at http://zeromq.org/area:licensing has been there for a while but it is hard to tell what projects it will apply to.

Is there any work in progress on moving this project to MPL?

daveyarwood:
Closes #326 and #327.


c-rack:
Thanks a lot!

c-rack:
By the way, the README still refers to the GPL.

daveyarwood:
Good catch -- I'll fix that on master.

trevorbernard:
Also, pom.xml has lgpl as it's license.

daveyarwood:
Another good catch -- fixing...

daveyarwood:
@trevorbernard It looks like there are numerous references to the LGPL as header comments in most (if not all?) of the source files. Example: https://github.com/zeromq/jeromq/blob/master/src/main/java/org/zeromq/ZThread.java#L1-L18

Should we remove these references to the license in the source files? Seems archaic in my opinion.

lbustelo:
@c-rack @trevorbernard @daveyarwood Great work!. This will open the doors for many more Apache.org projects using 0MQ.

PR #359, opened 2016-09-08
Remove headers from all source files

daveyarwood:
This PR is one way to complete #358 (migration from LGPL to MPLv2).

I realize this is probably controversial, but I think the most sustainable thing to do is to remove the headers from all source files. Most (but not all) of the source files currently have the same header that includes:

- the copyright (with outdated years, 2007-2014)
- the obvious statement that the file is part of the ZeroMQ project
- license information referring to LGPL

I would argue that having our source code organized in a GitHub repository and within a folder structure makes it clear that every source file is part of the ZeroMQ project, and that the README and LICENSE files in the root directory of the project make it obvious what our license and copyright are. In my humble opinion, it feels extraneous to include this information as a comment in every source file.

More importantly, it is not sustainable to have to update the copyright year of 210+ files every January 1, or to update the copyright/license information whenever it may change again in the future. So, I am advocating for removing the headers altogether.

Discussion welcome -- as a newcomer, I will defer to the rest of the team, and I will not be offended if we would prefer to update all the headers instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,4 +77,6 @@ Also please refer the [Wiki](https://github.com/zeromq/jeromq/wiki).
 
 ## Copying
 
+All source files are copyright © 2007-2016 contributors as noted in the AUTHORS file.
+
 Free use of this software is granted under the terms of the Mozilla Public License 2.0. For details see the file `LICENSE` included with the JeroMQ distribution.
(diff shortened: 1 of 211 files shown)


daveyarwood:
As I see it, we have three options:

- Merge this PR and do away with source headers altogether.
- Update all of the headers to refer to the MPLv2 license instead of LGPL.
- Update all of the headers to have a nonspecific reference to our license, pointing people to the `LICENSE` file for details.

c-rack:
I would follow the style of libzmq, which includes a license header in each source file. Using the [License Maven Plugin](http://www.mojohaus.org/license-maven-plugin/) should make this a very easy task. Just my 2 cents.

c-rack:
Sorry, forgot to merge this :)

PR #367, opened 2016-10-01
Use OpenJDK 8 for Travis

dermesser:
The patch in #366 uses the new Java Time API, which is only available in Java 8.

Alternatively (if you think that's the better way forward), I can try migrating my code to Java 7 (or lower).


dermesser:
I'll do it the other way :)

PR #368, opened 2016-10-01
Stop using java.time API

dermesser:
That API is only compatible with Java 8. (This replaces #367)


dermesser:
that build failure isn't my fault, at least :)

PR #374, opened 2016-10-20
A number of fixes, most ported from jzmq

bvella:
- fixed npe and mangling of existing indexes in ZMP.Poller (code synced with jzmq which does not have this issue)
- fixed windows bug in Signaler (reverted change introduced in 0.3.5)
- fixed failing tests on windows
- small changes to sync with jzmq

### DIFF
--- a/src/main/java/org/zeromq/ZMQException.java
+++ b/src/main/java/org/zeromq/ZMQException.java
@@ -4,17 +4,8 @@ import zmq.ZError;
 
 public class ZMQException extends RuntimeException
 {
-    public static class IOException extends RuntimeException
-    {
-        private static final long serialVersionUID = 8440355423370109164L;
 
-        public IOException(java.io.IOException cause)
-        {
-            super(cause);
-        }
-    }
-
-    private static final long serialVersionUID = 5957233088499712341L;
+    private static final long serialVersionUID = -978820750094924644L;
 
     private final int code;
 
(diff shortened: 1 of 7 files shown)


daveyarwood:
What's the reasoning for removing `hwm`, `getHWM` and `setHWM` in ZContext? Is `hwm` not used anywhere?

bvella:
no, there was no reference to the methods or the value anywhere. code compiles and tests run successfully

src/main/java/org/zeromq/ZMQException.java
```diff
@@ -4,17 +4,8 @@
...
-        private static final long serialVersionUID = 8440355423370109164L;
-
-        public IOException(java.io.IOException cause)
-        {
-            super(cause);
-        }
-    }
 
-    private static final long serialVersionUID = 5957233088499712341L;
+    private static final long serialVersionUID = -978820750094924644L;
```
daveyarwood: Just out of curiosity, what is the purpose of this value?

colini: Good explanation here:
http://stackoverflow.com/questions/285793/what-is-a-serialversionuid-and-why-should-i-use-it/285809

bvella: I compared jzmq and jeromq side by side and moved over a number of blocks that were different, including this value. To be fair, it should not make any difference, and I could as well have have left the value unchanged as i don't think anyone will be serializing this exception

daveyarwood: Ah, OK. Makes sense to me. It does seem like to keep JeroMQ consistent with jzmq, they two exceptions (which ought to be equivalent) should have the same serializable version ID.

Thanks for the explanation!

PR #386, opened 2016-11-28
Allow ZMQ.Context / ZContext to manage closing poll selectors

daveyarwood:
This PR picks up where #230 left off. The goal of this is to ensure that polling Selector resources are deallocated, in order to avoid "Too many open files" problems.

- I started with #230 and rebased on top of current master. There were no conflicts to resolve. I did have to update the ZMQ static method names in @fredoboulo 's new test file, as the method names had changed in master since he opened his PR.

- Per @trevorbernard 's suggestion, I un-reverted the fix in https://github.com/zeromq/jeromq/commit/4a19a453645d9c584345cb24868e9ff9b1fe02bd. I don't understand the full context, but it seems to me like this was fixed for a reason, and it's generally good for our code to be faithful to libzmq code. @trevorbernard also noted that shutdown would hang without that fix.

- There was a test that was removed for no apparent reason in that PR, so I added it back. The test passes.

- #230 added a null check in the PollSelector class. I've left that in, as it is better than just calling `selector.close()`.

- I did not include the `closePollSelector()` step that #230 added to the ZMQ.Context's `destroy` method. The implementation was to close the ThreadLocal `POLL_SELECTOR`, which @trevorbernard noted would not work in multithread use cases.

- I implemented a way for contexts to manage and deallocate poll selectors:
  - Both ZMQ.Context and ZContext now have a `List<Selector> selectors`.
  - Both types of context have a new method that will create a selector, adding it to the list of known selectors.
  - When the context is destroyed, we iterate through each selector, check if it is non-null, and if so, we close it, ensuring that resources are deallocated.
  - To make use of this, you must create the poller via `ZMQ.Context.poller()` / `ZMQ.Context.poller(int size)` or `ZContext.createPoller(int size)`, depending on which type of context you're using.

- Note that this PR does not affect ZPoller in any way. ZPoller does not use any of this code, and the tests are all passing.

### DIFF
--- a/src/main/java/org/zeromq/ZContext.java
+++ b/src/main/java/org/zeromq/ZContext.java
@@ -1,9 +1,7 @@
 package org.zeromq;
 
 import java.io.Closeable;
-import java.io.IOException;
 import java.nio.channels.Selector;
-import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
 
@@ -35,11 +33,6 @@ public class ZContext implements Closeable
     private List<Socket> sockets;
 
     /**
-     * List of selectors managed by this ZContext
-     */
-    private List<Selector> selectors;
-
-    /**
      * Number of io threads allocated to this context, default 1
      */
     private int ioThreads;
@@ -66,7 +59,6 @@ public class ZContext implements Closeable
     public ZContext(int ioThreads)
     {
         sockets = new CopyOnWriteArrayList<Socket>();
-        selectors = new ArrayList<Selector>();
         this.ioThreads = ioThreads;
         linger = 0;
         main = true;
@@ -86,18 +78,6 @@ public class ZContext implements Closeable
             socket.close();
         }
         sockets.clear();
-
-        for (Selector selector : selectors) {
-            if (selector != null) {
-                try {
-                    selector.close();
-                }
-                catch (IOException e) {
-                    throw new RuntimeException(e);
-                }
-            }
-        }
-        selectors.clear();
 
         // Only terminate context if we are on the main thread
         if (isMain() && context != null) {
@@ -149,15 +129,7 @@ public class ZContext implements Closeable
 
     public Selector createSelector()
     {
-        try {
-            // Create and register selector
-            Selector selector = Selector.open();
-            selectors.add(selector);
-            return selector;
-        }
-        catch (IOException e) {
-            throw new RuntimeException(e);
-        }
+        return getContext().selector();
     }
 
     public Poller createPoller(int size)
--- a/src/main/java/org/zeromq/ZContext.java
+++ b/src/main/java/org/zeromq/ZContext.java
@@ -1,10 +1,14 @@
 package org.zeromq;
 
 import java.io.Closeable;
+import java.io.IOException;
+import java.nio.channels.Selector;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
 
 import org.zeromq.ZMQ.Context;
+import org.zeromq.ZMQ.Poller;
 import org.zeromq.ZMQ.Socket;
 
 import zmq.ZError;
@@ -29,6 +33,11 @@ public class ZContext implements Closeable
      * List of sockets managed by this ZContext
      */
     private List<Socket> sockets;
+
+    /**
+     * List of selectors managed by this ZContext
+     */
+    private List<Selector> selectors;
 
     /**
      * Number of io threads allocated to this context, default 1
@@ -57,6 +66,7 @@ public class ZContext implements Closeable
     public ZContext(int ioThreads)
     {
         sockets = new CopyOnWriteArrayList<Socket>();
+        selectors = new ArrayList<Selector>();
         this.ioThreads = ioThreads;
         linger = 0;
         main = true;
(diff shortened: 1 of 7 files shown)


daveyarwood:
NB: I don't think the Travis tests failing on this branch are related to the contents of the PR. There are existing issues causing Travis builds to fail.

fredoboulo:
I have some comments about your PR which indeed improves the behaviour. I will add them a bit later at the end of my work day.

src/main/java/org/zeromq/ZContext.java
```diff
@@ -31,6 +35,11 @@
     private List<Socket> sockets;
 
     /**
+     * List of selectors managed by this ZContext
+     */
+    private List<Selector> selectors;
+
+    /**
```
fredoboulo: It would be possible to handle the list of selectors inside the Ctx class, reducing the amount of duplicated code and the 3 layers of context (Ctx, Context, ZContext) would benefit of the mechanism.

src/main/java/zmq/Ctx.java
```diff
@@ -127,6 +134,13 @@ private void destroy() throws IOException
             it.close();
         }
 
+        for (Selector selector : selectors) {
+            if (selector != null) {
```
fredoboulo: The null checking may not be necessary here.

src/main/java/zmq/ZMQ.java
```diff
@@ -822,14 +822,23 @@ public Selector get()
...
+                }
+                catch (IOException e) {
+                    e.printStackTrace();
+                }
+            }
+        }
+
         @Override
         public void finalize()
         {
```
fredoboulo: ATM, this is one of the place of the memory leaks. The code should never rely on finalize() being called to release the resource.

fredoboulo:
Displaying the usages (or opening the call hierachy) on my IDE for the Selector.open() method, I find that this is called in several places throughout the library. My original intent was to concentrate all these openings (and closings) in one place in the code to get a better management of these resources, but for that I would have liked to make it step by step. PR #230 was the first step.

src/main/java/zmq/ZMQ.java
trevorbernard: I agree, there are a lot of ramifications involving the use of `finalize` that most people don't understand. I would go so far as to say that you should never override it unless you know exactly what your doing.

src/main/java/org/zeromq/ZContext.java
daveyarwood: That's a good idea. ZContext actually has a Context (not a Ctx directly), but Context has a Ctx, so by extension, every ZContext has a Ctx. So, we should be able to put the list of Selectors at the "lowest level" Ctx and avoid duplicating code.

I'll take a stab at this soon when I have a minute.

src/main/java/zmq/Ctx.java
daveyarwood: That thought occurred to me as well... do you think it's worth leaving the null check in there, just in case? I don't know the full context, but I haven't seen any code that suggests a selector in the list might be null.

daveyarwood:
@fredoboulo Thanks for the review! I'll make a revision soon.

re: the `finalize` code, I didn't see any way to remove it entirely, as the way we have Pollers set up, they can exist outside of contexts, and rely on the ThreadLocal variable to manage their Selectors.

Example code from mspoller.java:

```java
//  Initialize poll set
ZMQ.Poller items = new ZMQ.Poller (2);
items.register(receiver, ZMQ.Poller.POLLIN);
items.register(subscriber, ZMQ.Poller.POLLIN);
```

This constructor of `ZMQ.Poller` [passes in `null` as the Context](https://github.com/zeromq/jeromq/blob/master/src/main/java/org/zeromq/ZMQ.java#L1482-L1485), which makes it impossible for the context to manage the poller's selector.

I think the best thing to do would be to remove the ability to create a Poller in any way except through the context like in the code added by this PR, and then remove all the dead code that depends on the ThreadLocal selectors.

That would be a breaking change, though, so I figured it would be best to leave in the code that uses the ThreadLocal selectors to avoid breaking users' code.

Maybe I could mark the old methods as deprecated in favor of the new context-managed Poller constructors?

src/main/java/zmq/Ctx.java
colini: If there should never be a null entry in the list, then another approach is to enforce the null check when a selector is added and throw NPE at that time.

fredoboulo: It does no harm to leave the checking here. In #230, the null checking is to avoid getting a NPE, as the selector is a member of the class that can be nullified (basically calling 2 or more times close() should be done without side-effects).
The call to Selector.open() is the only way to add one into the list, so if this method is trusted to always deliver an instance, there is no need to do so. As you stated @colini, an assert(selector != null) should suffice.
Here there is the liberty to choose which defensive programming you want to apply, in one case the path is to close and release resources, in the other it is about allocating the resources, both stages where performance may not be the most important point.

src/main/java/zmq/ZMQ.java
fredoboulo: This PollSelector class is not part of libzmq, as far as I can see. ATM, it was a workaround to comply with the original library, so this code may be refactored, or rather its use enforced to provide a correct behaviour.

src/main/java/zmq/Ctx.java
daveyarwood: I've opted to add an `assert(selector != null)` upon opening a Selector, and also checking for null before closing them, just in case, since there is some possibility that calling code could be setting a selector to `null`.

src/main/java/zmq/ZMQ.java
daveyarwood: I think I'll create a separate PR building on top of this one that removes the PollSelector completely and forces the user to only create Selectors (or Pollers) via a context.

This would be a breaking change, but I think it's a step in the right direction. I do think it should be a separate PR, though, so that this PR can be strictly non-breaking changes.

daveyarwood:
I've made a few additional commits to address feedback received thus far.

I think this is ready to merge, but any additional feedback is appreciated!

PR #387, opened 2016-12-04
Remove PollSelector

daveyarwood:
This is a WIP PR building from #386.

In reviewing #386, @fredoboulo noted that:

> This PollSelector class is not part of libzmq, as far as I can see. ATM, it was a workaround to comply with the original library, so this code may be refactored, or rather its use enforced to provide a correct behaviour.

This PR does just that by removing the PollSelector class altogether, along with all code using it.

The main consequence of this is that it enforces that you must use a ZMQ context in order to create a poller. The idea is that a poller should not exist outside of a ZMQ context; the context needs to be able to shut down its pollers and free the selector resources, which is only possible if every poller belongs to a context.


daveyarwood:
The "WIP" part of this PR is that the ZLoop tests are failing, and I'm not sure why. Something about passing in a context when constructing a ZLoop seems to be causing its pollers not to work correctly. I could use some help with this, it's a real head-scratcher.

fredoboulo:
Wouah, that's a huge commit you want to merge here. I will look for a way to make smaller steps to achieve the result, please give me some time (and remind me if it is too long)

daveyarwood:
@fredoboulo The changes to the library itself in that commit are small -- most of the commit is updating all of the test files to create pollers via a context rather than as "independent" pollers not associated with a context.

src/main/java/org/zeromq/ZMQ.java
```diff
@@ -1526,13 +1508,11 @@ public Poller(int size)
...
         protected Poller(Context context, int size)
         {
-            if (context != null) {
-                this.context = context;
-            }
-
-            items = new PollItem[size];
-            timeout = -1L;
-            next = 0;
+            this.context = context;
```
fredoboulo: Add assert(context != null) would be nice checking and fit with the current code

src/main/java/org/zeromq/ZMQ.java
```diff
@@ -1526,13 +1508,11 @@ public Poller(int size)
...
+            selector     = context.selector();
+            items        = new PollItem[size];
```
fredoboulo: Add assert(selector != null) would be nice checking and fit with the current code

fredoboulo:
Got the ZLoop problem: the context was not initialized.

Below is a diff to fix the situation.

diff --git a/src/main/java/org/zeromq/ZLoop.java b/src/main/java/org/zeromq/ZLoop.java
index d9efbfb..a3de64c 100644
--- a/src/main/java/org/zeromq/ZLoop.java
+++ b/src/main/java/org/zeromq/ZLoop.java
@@ -59,7 +59,7 @@ public class ZLoop

     }

-    private Context context;                    //  Context managing the pollers.
+    private final Context context;              //  Context managing the pollers.
     private final List<SPoller> pollers;        //  List of poll items
     private final List<STimer> timers;          //  List of timers
     private int pollSize;                       //  Size of poll set
@@ -74,7 +74,7 @@ public class ZLoop
     {
         assert (context != null);

-        context   = context;
+        this.context   = context;
         pollers   = new ArrayList<SPoller>();
         timers    = new ArrayList<STimer>();
         zombies   = new ArrayList<Object>();

src/main/java/org/zeromq/ZLoop.java
```diff
@@ -58,6 +59,7 @@ public STimer(int delay, int times, IZLoopHandler handler,
 
     }
 
+    private Context context;                    //  Context managing the pollers.
```
fredoboulo: private **final** Context context;

src/main/java/org/zeromq/ZLoop.java
```diff
@@ -68,14 +70,22 @@ public STimer(int delay, int times, IZLoopHandler handler,
...
 
-    public ZLoop()
+    public ZLoop(Context context)
     {
-        pollers = new ArrayList<SPoller>();
-        timers = new ArrayList<STimer>();
-        zombies = new ArrayList<Object>();
+        assert (context != null);
+
+        context   = context;
```
fredoboulo: **this.**context = context;

fredoboulo:
Got the ZLoop problem: the context was not initialized.
I wrote the fix in the comments of the PR

daveyarwood:
Aha -- good catch! The tests are passing for me now 🎉

src/main/java/org/zeromq/ZLoop.java
```diff
@@ -242,13 +252,17 @@ public int start()
...
             }
             long wait = ticklessTimer();
 
-            rc = pollset.poll(wait);
+            if (pollset != null) {
+              rc = pollset.poll(wait);
+            }
 
-            if (rc == -1) {
+            if (context == null || rc == -1) {
```
fredoboulo: the null checking should be irrelevant now

src/test/java/org/zeromq/TestDisconnectInproc.java
```diff
@@ -0,0 +1,94 @@
+package zmq;
```
fredoboulo: I missed that small one. Is there any reason why the test was moved from package zmq? Right now, it got namespace problem, and with ZMQ.ZMQ_X socket types...

daveyarwood:
@fredoboulo Do you think this is good to merge?

daveyarwood:
(rebased on top of master)

daveyarwood:
There don't seem to be any objections, so I'm going to go ahead and merge it.

This is a breaking change, so I think the next release of JeroMQ should bump the minor version and be 0.4.0.

PR #389, opened 2016-12-05
Ensure that the tests always use ports that aren't already in use

### LINKED ISSUE #385: Tests intermittently fail because of hard-coded ports that may not be available
Example test run:

```
Running org.zeromq.ZBeaconTest
Exception in thread "Thread-26" java.lang.RuntimeException: java.net.BindException: Can't assign requested address
        at org.zeromq.ZBeacon$BroadcastClient.run(ZBeacon.java:137)
Caused by: java.net.BindException: Can't assign requested address
```

The corresponding code in ZBeaconTest.java includes this line:

```java
ZBeacon zbeacon = new ZBeacon("255.255.255.255", 5670, beacon, false);
```

Apparently I already have something running on 5670, so the test failed.

Action needed:
- Helper function that finds an open port, e.g. initialize a ServerSocket with port "0" so it finds an open port, use `socket.getLocalPort()` to get the port number, close the port, and return the port number.
- Go through the tests and find all the places where we use hard-coded ports, and use the helper function instead to ensure that the tests always use ports that aren't already in use.

daveyarwood:
This makes the tests more deterministic by not using hard-coded ports that might already be in use by another process on your computer. I've noticed tests failing both on my machine and in Travis builds for this reason. Instead, we can use the ServerSocket constructor (wrapped up in a new util method, `findOpenPort`) to find ports that are guaranteed to be available.

Fixes #385.

### DIFF
--- a/src/test/java/org/zeromq/ZBeaconTest.java
+++ b/src/test/java/org/zeromq/ZBeaconTest.java
@@ -1,5 +1,6 @@
 package org.zeromq;
 
+import java.io.IOException;
 import java.net.InetAddress;
 
 import org.junit.Test;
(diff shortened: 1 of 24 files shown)


src/test/java/org/zeromq/ZBeaconTest.java
```diff
@@ -12,12 +13,13 @@
...
     @Test
-    public void test() throws InterruptedException
+    public void test() throws InterruptedException, IOException
     {
         final CountDownLatch latch = new CountDownLatch(1);
         byte[] beacon = new byte[] { 'H', 'Y', 'D', 'R', 'A', 0x01, 0x12, 0x34 };
         byte[] prefix = new byte[] { 'H', 'Y', 'D', 'R', 'A', 0x01 };
-        ZBeacon zbeacon = new ZBeacon("255.255.255.255", 5670, beacon, false);
+        int port = Utils.findOpenPort();
+        ZBeacon zbeacon = new ZBeacon("127.0.0.1", port, beacon, false);
```
colini: Is 127.0.0.1 preferable here over 255.255.255.255 ?

PR #419, opened 2017-04-16
Fix erroneous javadoc references in ZMQ

thrau:
These are very minor non-functional changes that correct a few erroneous javadoc references to methods that were previously moved or had their signature changed.


c-rack:
Thanks!

PR #421, opened 2017-04-18
change of API for org.zeromq.ZMQ to be consistent with existing

fredoboulo:


fredoboulo:
Thanks to @thrau, I could see the weaknesses of my first changes to org.zeromq.ZMQ. 
I updated the getters and setters of socket options, together with comments from http://api.zeromq.org/4-1:zmq-getsockopt. Should be better, please tell me otherwise.

thrau:
just a minor thing (which was what my change addressed ;-) ): in the javadoc to `public boolean getIPv4Only()`, there is a link that now points to `isIpv6`. should be reverted to `isIPv6` in order for a correct link to be generated in the javadoc.

Issue #451, opened 2017-08-10
Support for Android API Level < 24?


skipe777:
The library (version 0.4.2) seems to leverage java.net.StandardProtocolFamily and java.net.StandardSocketOptions which are only available at API level 24 and greater. Is there a version of JeroMQ that operates on a lower API level then this (with Curve support)?

fredoboulo:
I just made a PR to fix that. Could you test it with Android environment to be sure that no other point needs to be addressed for a successful compile?

skipe777:
A bit more progress, still complaining about 24 or greater APIs:

`java.lang.NoClassDefFoundError: Failed resolution of: Ljava/nio/channels/NetworkChannel;
                                                                                at zmq.io.net.tcp.TcpListener.setAddress(TcpListener.java:186)
                                                                                at zmq.SocketBase.bind(SocketBase.java:363)
                                                                                at org.zeromq.ZMQ$Socket.bind(ZMQ.java:2333)`

skipe777:
It doesn't make it to that call, it stops at setReuseAddress():

[...]

PR #457, opened 2017-08-23
Make jeromq usable with Android API < 24

fredoboulo:
For #451

### DIFF
--- a/src/main/java/zmq/io/net/tcp/TcpUtils.java
+++ b/src/main/java/zmq/io/net/tcp/TcpUtils.java
@@ -1,11 +1,13 @@
 package zmq.io.net.tcp;
 
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


ipechorin:
testPushPull500 (org.zeromq.TestPushPullThreadedTcp)  fails, @fredoboulo can you please have a look at this. I'll revert the change for now.

fredoboulo:
Yes it is failing not systematically for quite some time now (TestPushPullThreadedTcp and TestZProxy are painful for that), this commit is not the source of the problem.
I'm preparing a PR to fix that, the thing is that most of the time, the CI build of the PR succeeds then the master CI build fails, so it is not very convenient to assess

fredoboulo:
Please merge it back when you feel the time is right

Issue #447, opened 2017-07-20
Divide by zero exception.


madrzak:
Divide by zero occurs very occasionally on the subscriber in the PUB SUB pattern. Seems to occur more frequently when internet connection is weak. 

Code (crashes on the second line)
`String t = subscriber.recvStr();
  String msg = subscriber.recvStr();`

Exception:
java.lang.ArithmeticException: divide by zero
	at zmq.FQ.recvPipe(FQ.java:88)
	at zmq.FQ.recv(FQ.java:67)
	at zmq.XSub.xrecv(XSub.java:165)
	at zmq.SocketBase.recv(SocketBase.java:703)
	at org.zeromq.ZMQ$Socket.recv(ZMQ.java:1377)
	at org.zeromq.ZMQ$Socket.recvStr(ZMQ.java:1448)
	at org.zeromq.ZMQ$Socket.recvStr(ZMQ.java:1438)

colini:
Which version of jeromq?

madrzak:
'org.zeromq:jeromq:0.4.0'

daveyarwood:
This line looks to be the culprit: https://github.com/zeromq/jeromq/blob/v0.4.0/src/main/java/zmq/FQ.java#L88

colini:
Looking at `recvPipe()` I don't see how active could be 0 at line 88.  @madrzak are you sure your socket is only accessed by the thread that created it?  0MQ sockets are not thread safe.

fredoboulo:
Code is quite the same in libzmq: https://github.com/zeromq/libzmq/blob/master/src/fq.cpp#L110
Worth to know is that a fix was made in LB to avoid a divide by zero Exception: https://github.com/zeromq/libzmq/blob/master/src/lb.cpp#L141, and that this fix already exists in Jeromq: https://github.com/zeromq/jeromq/blob/master/src/main/java/zmq/socket/LB.java#L117

A fix can also be made for that to avoid the exception, but if the socket is accessed from multiple threads, you will get unpredictable behaviour (with no crash). Could you provide a minimal test case that pinpoints the problem?

daveyarwood:
> A fix can also be made for that to avoid the exception, but if the socket is accessed from multiple threads, you will get unpredictable behaviour (with no crash).

In that case, it seems like we should be catching the divide by zero exception and wrapping it with a more specific ZMQ exception.

fredoboulo:
@madrzak May you provide us compilable code demonstrating the exception?

[...]

PR #465, opened 2017-08-25
Small changes

### LINKED ISSUE #446: ZMQ.Socket class is final
Hello

While migrating from 0.3.6 to the latest 0.4.2 I have noticed that ZMQ.Socket class has been changed to be final so I can't extend it any more by my custom Socket class. What was the reason behind that change? As far as I understand jeromq tries to mimic the jzmq API and it seems that making ZMQ.Socket class final breaks this contract.

Cheers,
Serguei

fredoboulo:
Assert on FQ to assess single-threading behavior on receive (related to #447)
ZMQ.Socket class is no longer final (fixes #446)
added jacoco coverage as a proposal
Small javadoc changes
Handle interrupt by close in ZBeacon (quite rare occurrences)

### DIFF
--- a/src/main/java/org/zeromq/ZActor.java
+++ b/src/main/java/org/zeromq/ZActor.java
@@ -39,8 +39,8 @@ import org.zeromq.ZPoller.EventsHandler;
  * <br/>
  * The provided {@link Actor} is living on the Plateau.
  * <br/>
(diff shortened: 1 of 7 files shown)


daveyarwood:
Looks good to me :+1:

fredoboulo:
VM of travis CI crashed, I force a recheck

fredoboulo:
Jacoco makes Travis crash, too bad :/ i removed that from the commit

PR #469, opened 2017-09-05
Typos and warning removals

fredoboulo:
Better styling: removed unnecessary keywords and var assignments
Fixed error in comparison of byte[] in Mechanism
Handled possibility of null received message in ZSocket
Fixed return status of ZMQ#setHWM

Some warnings were about dead branches of code, most critical ones are in Signaler and Mechanism.


trevorbernard:
`static` inner classes have special meaning, by making it `static`, it causes the class not to have a reference to the parent class.

fredoboulo:
Done.

PR #471, opened 2017-09-06
lib now usable with Android API 10 and above.

### LINKED ISSUE #451: Support for Android API Level < 24?

fredoboulo:
Fixes #451

### DIFF
--- a/src/main/java/zmq/io/Metadata.java
+++ b/src/main/java/zmq/io/Metadata.java
@@ -1,6 +1,5 @@
 package zmq.io;
 
-import java.util.Objects;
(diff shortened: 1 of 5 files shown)


Itaydal:
Hi @fredoboulo 

My tests on Android are passing, hope I haven't messed anything..

Thanks a lot :)

Issue #472, opened 2017-09-07
Enforce code resilience and usage


fredoboulo:
These last weeks, bugs impacting major features were fixed.

The big thing for me is that these ultra important major features were absolutely never tested: downgrading capabilities, socket reconnection, Android compatibility. I take the responsibilities of my mistakes, and I would like to go forward on that matter and be constructive about it.

I propose to enforce testing scenarii so no commit will provoke a regression of that level.

- Downgrading capabilities:
We can always introduce some option to force protocol version usage, that will allow to test protocol compatibilities within Jeromq. Not the best, but still  a step forward. I would prefer environments set with different versions of libzmq, so testing can connect and exchange messages with them.

- Socket reconnection:
I have currently no idea how we can test that at end-to-end level. Any suggestion would be welcome.

- Android compatibility
Right now the code is compatible with Android API level 10. The thing is that JDKs are not fully implemented depending on the versions (some classes, some methods are missing depending on this level). Is it possible to setup Travis so we can at least run a compilation of the project on Android environment ([distribution of devices is here](https://developer.android.com/about/dashboards/index.html))?

trevorbernard:
I completely agree and endorse this effort. We're certainly blind in the areas you mentioned.

daveyarwood:
I also think this is a great idea, very worthwhile.

Re: Travis CI for Android, it looks like that ought to be fairly easy to set it up so that we run the build and tests on a variety of Android SDK versions: https://docs.travis-ci.com/user/languages/android/

It seems like it might be a good idea to set up automated end-to-end testing as a separate repo in the zeromq org, since that would be testing not only JeroMQ but other flavors of libzmq. If some of those tests fail, it might not necessarily be an indicative of a problem with JeroMQ, so we wouldn't want those failures to block the development of JeroMQ.

fredoboulo:
That sounds promising!

For downgrading capabilities, I was thinking about a set of well-known and stable releases of libzmq, that implement flawlessly the different protocols. So if there is a miscommunication, it can be certified that it comes from the tested side.

daveyarwood:
Ah, that makes sense. Would it be fair to consider libzmq (using the C library directly) the "gold standard" against we measure other ZMQ implementations?

fredoboulo:
It's fair enough for me. 
We can always go in full blossomed matrix mode at a latter step, but I would be more than pleased having  comms tested with libzmq in all protocol versions and formats (v0 to v3).

[...]

PR #481, opened 2017-09-13
Add Coveralls Maven plugin and Travis CI integration

daveyarwood:
For #472.

I tried running `mvn clean test jacoco:report coveralls:report -P coverage` locally and it works except for the Coveralls part. I did some reading on this, and it looks like if we ever want to run the Coveralls submit task locally, we need to include `-DrepoToken=coverallsrepotokengoeshere`, per the Coveralls Maven plugin README.

I wasn't able to test that part because I don't have admin access to the JeroMQ repo, therefore I cannot see the repo token in the Coveralls UI, [as noted here](https://github.com/lemurheavy/coveralls-public/issues/152#issuecomment-33510656).

On the other hand, Travis should be able to start submitting coverage reports to Coveralls out-of-the-box, from what I've read, so hopefully this will just work™ when this PR is merged and Travis runs on the master branch.

It may or may not work when Travis runs it as part of this PR, as I think Coveralls is set up to not run in PRs, otherwise anybody could alter our published coverage percentage based on the content of their PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,4 +3,5 @@ language: java
 jdk:
   - openjdk7
   - oraclejdk8
-
+after_success:
+  - mvn clean test jacoco:report coveralls:report -P coverage
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
 Pure Java implementation of libzmq (http://zeromq.org).
 
 [![Build Status](https://travis-ci.org/zeromq/jeromq.png)](https://travis-ci.org/zeromq/jeromq)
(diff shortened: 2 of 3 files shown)


daveyarwood:
Note: if this causes the Travis build to fail on the `after_success` hook, I can alter it so that Travis doesn't try to do the Coveralls part if it's building a PR.

daveyarwood:
Cool, it worked, and it even commented on the PR! This is really awesome.

And it looks like the overall badge says "coverage unknown":  [Coverage Status](https://coveralls.io/github/zeromq/jeromq?branch=master)

which is good because we don't want PR builds influencing our coverage percentage -- we want the coverage percentage for the master branch.

sigiesec:
Cool :+1:

fredoboulo:
Awesome!

Issue #198, opened 2014-10-19
Errors when rapidly opening/closing sockets and contexts


augustl:
I'm attempting to create an instance of my ZMQ stack for each unit test in my test suite. But I get problems when I rapidly open/close sockets and contexts.

Some times I get this:

```
Exception in thread "reaper-1" java.lang.AssertionError
    at zmq.Mailbox.recv(Mailbox.java:114)
    at zmq.SocketBase.process_commands(SocketBase.java:830)
    at zmq.SocketBase.in_event(SocketBase.java:927)
    at zmq.Poller.run(Poller.java:237)
    at java.lang.Thread.run(Thread.java:745)
```

Other times it hangs silently at ctx.term();

I've created a maven project with a single junit test that demonstrates the issue: https://github.com/augustl/jeromq-issue

For the record, the test case is:

``` java
public class ExampleText {
    @Test
    public void demonstrateIssue() {
        for (int i = 0; i < 50; i++) {
            performTest();
        }
    }

    private void performTest() {
        Context ctx = ZMQ.context(1);
        Socket recvMsgSock = ctx.socket(ZMQ.PULL);
        recvMsgSock.bind("tcp://*:5115");
        Socket processMsgSock = ctx.socket(ZMQ.PUSH);
        processMsgSock.bind("inproc://process-msg");

        List<Socket> workerSocks = new ArrayList<Socket>();
        for (int i = 0; i < 5; i++) {
            Socket workerSock = ctx.socket(ZMQ.PULL);
            workerSock.connect("inproc://process-msg");
            workerSocks.add(workerSock);
        }

        Thread proxyThr = new Thread(new ZMQQueue(ctx, recvMsgSock, processMsgSock));
        proxyThr.setName("Proxy thr");
        proxyThr.start();

        for (final Socket workerSock : workerSocks) {
            Thread workerThr = new Thread(new Runnable() {
                @Override
                public void run() {
                    try {
                        while (true) {
                            byte[] msg = workerSock.recv();
                            // Process the msg!
                        }
                    } catch (Exception e) {

                    }
                }
            });
            workerThr.setName("A worker thread");
            workerThr.start();
        }

        try {
            Thread.sleep(100);
        } catch (InterruptedException e) {
            throw new RuntimeException(e);
        }

        System.out.println("Closing now");


        recvMsgSock.close();
        processMsgSock.close();

        for (Socket workerSock : workerSocks) {
            workerSock.close();
        }

        ctx.term();
[...]

PR #493, opened 2017-10-15
Add a (failing) test for issue #198

daveyarwood:
In the interest of working toward a solution to #198, I copied @augustl's code into a new test file and patched it up a little bit to work with the current state of JeroMQ.

I basically just replaced the `ZMQ.context(1)` with a `new ZContext()` and replaced the context methods with the corresponding methods from ZContext (e.g. `ctx.socket` became `ctx.createSocket`, `ctx.term` became `ctx.destroy`).

I also removed the lines near the end that manually close the sockets, as this is something that the ZContext does for you when it is terminated.

Rather than use the same port every time, I used `Utils.findOpenPort` to use a new available port on every run, so we can at least conclude that the issue isn't something to do with reusing the same port.

I was able to reproduce the issue easily -- depending on the test run, the process would hang on termination of the context either on the first run, or it would get a little farther than that and hang around the third run.

### DIFF
--- a/src/test/java/org/zeromq/TestRapidOpenCloseSocket.java
+++ b/src/test/java/org/zeromq/TestRapidOpenCloseSocket.java
@@ -10,7 +10,7 @@ import org.zeromq.ZMQ.Socket;
 // For issue: https://github.com/zeromq/jeromq/issues/198
 public class TestRapidOpenCloseSocket
 {
-  @Test
+  @Test(timeout = 10000)
   public void testRapidOpenCloseSocket() throws Exception
   {
     for (int i = 0; i < 50; i++) {
--- a/src/test/java/org/zeromq/TestRapidOpenCloseSocket.java
+++ b/src/test/java/org/zeromq/TestRapidOpenCloseSocket.java
@@ -3,6 +3,7 @@ package org.zeromq;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.junit.Ignore;
 import org.junit.Test;
 
 import org.zeromq.ZMQ.Socket;
@@ -10,7 +11,8 @@ import org.zeromq.ZMQ.Socket;
 // For issue: https://github.com/zeromq/jeromq/issues/198
 public class TestRapidOpenCloseSocket
 {
-  @Test(timeout = 10000)
+  @Test
+  @Ignore
   public void testRapidOpenCloseSocket() throws Exception
   {
     for (int i = 0; i < 50; i++) {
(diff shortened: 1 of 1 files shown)


src/test/java/org/zeromq/TestRapidOpenCloseSocket.java
```diff
@@ -0,0 +1,75 @@
...
+
+import org.junit.Test;
+
+import org.zeromq.ZMQ.Socket;
+
+// For issue: https://github.com/zeromq/jeromq/issues/198
+public class TestRapidOpenCloseSocket
+{
+  @Test
+  public void testRapidOpenCloseSocket() throws Exception
```
fredoboulo: Please add @Ignore, so the failing test is not blocking CI. Adding timeout on test could also be useful.

daveyarwood:
Good call -- we don't want this test to make it so that we can't run the entire test battery.

It turns out that JUnit supports [setting a timeout on tests]() -- I learned something today.

I've added a 10,000ms timeout so that the test runs in a thread and if it takes longer than 10 seconds to complete, the thread is killed, the test is counted as an error, and we can move on with running the other tests:

```
Running org.zeromq.TestRapidOpenCloseSocket
Closing now
Successfully closed
Closing now
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 10.108 sec <<< FAILURE! - in org.zeromq.TestRapidOpenCloseSocket
testRapidOpenCloseSocket(org.zeromq.TestRapidOpenCloseSocket)  Time elapsed: 10.023 sec  <<< ERROR!
java.lang.Exception: test timed out after 10000 milliseconds
```

src/test/java/org/zeromq/TestRapidOpenCloseSocket.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.List;
+
+import org.junit.Test;
+
+import org.zeromq.ZMQ.Socket;
+
+// For issue: https://github.com/zeromq/jeromq/issues/198
+public class TestRapidOpenCloseSocket
+{
+  @Test(timeout = 10000)
```
fredoboulo: Definitely put @Ignore. This is a failing test as there is no fix available as of now, and this is a situation that is present since... a loooong time, so it is mostly to document behaviour so we can have a baseline.
If this is merged, it will provoke systematic fail on every subsequent PRs, preventing a better visibility of the status of the lib.
When this is fixed, let's just remove the annotation and let's make a toast for that :)

daveyarwood: Sounds reasonable enough -- done.

fredoboulo:
Hi, I'm wondering why this PR is stalled for so long. Any reason ?

daveyarwood:
@fredoboulo I think it's ready to merge, if you agree. Note that the test is marked `@ignore` for now -- I think your suggestion was that we un-ignore it when we have a fix in place.

Issue #488, opened 2017-09-29
Polling a DEALER socket sometimes causes IndexOutOfBoundsException


daveyarwood:
The context is that I'm writing a [Clojure JeroMQ wrapper library](https://github.com/daveyarwood/ezzmq) and translating the zguide examples as I go.

When I translated the multi-threaded server example (a single process with a ROUTER/DEALER proxy and 5 spawned worker threads running a REP server), ran it, and interrupted it via Ctrl-C, I found that the worker processes caught ETERM and exited successfully, but the server failed to shut down and blocked the process from exiting.

Occasionally, I got this stacktrace:

```
...
                                            ezzmq.poll/poll                     poll.clj:   71
                                            ezzmq.poll/poll                     poll.clj:   85
                                                        ...            
                                 org.zeromq.ZMQ$Poller.poll                     ZMQ.java: 2903
                                 org.zeromq.ZMQ$Poller.poll                     ZMQ.java: 2939
                                               zmq.ZMQ.poll                     ZMQ.java:  709
                                 zmq.poll.PollItem.readyOps                PollItem.java:  115
                                zmq.SocketBase.getSocketOpt              SocketBase.java:  257
                                      zmq.SocketBase.hasOut              SocketBase.java:  878
                           zmq.socket.reqrep.Dealer.xhasOut                  Dealer.java:   83
                                       zmq.socket.LB.hasOut                      LB.java:  135
                                    java.util.ArrayList.get               ArrayList.java:  429
                             java.util.ArrayList.rangeCheck               ArrayList.java:  653
java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
         clojure.lang.ExceptionInfo: Index: 0, Size: 0
    line: 73
```

I traced it back to [this line in LB.java](https://github.com/zeromq/jeromq/blob/master/src/main/java/zmq/socket/LB.java#L135), which calls ArrayList.get on an index that is out of bounds. My guess is that the worker processes exited, so `pipes` is now of length 0, but we try to get `pipes[0]`.

I don't have much context on this part of the code, but I have two observations:

1. Is there a bug here?

2. If it's not a bug, since it's possible to get into this predicament, should we at least catch the exception and throw a more descriptive one? (or perhaps handle it more gracefully, like `return false` or something?)

fredoboulo:
Difficult to say now and here without more infos, and I am not familiar with clojure, maybe someone else would provide better answers.

The usual (you know the drill)
[...]

PR #502, opened 2017-11-16
 Issue #488 - Make ZContext more like zctx

trevorbernard:


src/main/java/org/zeromq/ZContext.java
```diff
@@ -116,10 +145,14 @@ public void destroySocket(Socket s)
...
-            s.setLinger(linger);
-            s.close();
+        s.setLinger(linger);
+        s.close();
+        try {
+            mutex.lock();
+            this.sockets.remove(s);
+        }
+        finally {
+            mutex.unlock();
```
daveyarwood: This part looks especially good to me.

The way the code is now, it seems like it's possible for a socket to sometimes not be closed, e.g. if it wasn't able to be removed because maybe it was removed in another thread.

The way you have it, we definitely close the socket, and then we remove it (within a locked mutex context, which is good), which seems safer.

trevorbernard:
I basically brought it up to date with zctx before it was deprecated and removed. Using explicit locks is easier to reason about than using synchronization, etc. I remember sending you a note about this but totally forgot I had it in my fork.

daveyarwood:
Oh, interesting... So does that mean, if we're following the development path of libzmq, that ZContext will eventually be deprecated?

Is ZSocket (i.e. an invisible, managed context) the next step?

trevorbernard:
That was Pieter's vision for czmq

daveyarwood:
OK, that makes sense. I'm interested in working in that direction, too. Ideally the context would just be an implementation detail that application code never (or at least rarely) has to deal with.

One step at a time, though :)

daveyarwood:
👍 on this PR, BTW. Shall I merge it?

trevorbernard:
Yes please. Also, any PR that satisfies the C4 should always be merged.

PR #510, opened 2017-12-20
Avoid using system time to measure elapsed time

pepijnve:
I ran into a situation today where Poller#poll was blocked for a long time even though a timeout value of 1 second had been passed to poll. I attached a debugger and captured the state displayed in the attached screenshot. As you can see `waitMillis` has a value of just over an hour even though `timeout` is set to 1000.

This can happen because `poll` uses `currentTimeMillis` values to measure elapsed time without taking into account that the system clock may be changed in the meantime.

To avoid this I've updated `poll` to use `nanoTime` instead. This way elapsed time is measured correctly regardless of system time changes.

screen shot 2017-12-20 at 12 08 12

### DIFF
--- a/src/main/java/zmq/ZMQ.java
+++ b/src/main/java/zmq/ZMQ.java
@@ -695,7 +695,7 @@ public class ZMQ
                 waitMillis = -1L;
             }
             else {
-                waitMillis = TimeUnit.MICROSECONDS.toMillis(end - now);
+                waitMillis = TimeUnit.NANOSECONDS.toMillis(end - now);
                 if (waitMillis == 0) {
                     waitMillis = 1L;
                 }
@@ -754,8 +754,8 @@ public class ZMQ
             //  first pass have taken negligible time). We also compute the time
             //  when the polling should time out.
             if (firstPass) {
-                now = Clock.nowUS();
-                end = now + TimeUnit.MILLISECONDS.toMicros(timeout);
+                now = Clock.nowNS();
+                end = now + TimeUnit.MILLISECONDS.toNanos(timeout);
                 if (now == end) {
                     break;
                 }
@@ -764,7 +764,7 @@ public class ZMQ
             }
 
             //  Find out whether timeout have expired.
-            now = Clock.nowUS();
+            now = Clock.nowNS();
             if (now >= end) {
                 break;
             }
--- a/src/main/java/zmq/ZMQ.java
+++ b/src/main/java/zmq/ZMQ.java
@@ -695,7 +695,10 @@ public class ZMQ
                 waitMillis = -1L;
             }
             else {
-                waitMillis = end - now;
+                waitMillis = TimeUnit.MICROSECONDS.toMillis(end - now);
+                if (waitMillis == 0) {
+                    waitMillis = 1L;
+                }
             }
 
             //  Wait for events.
@@ -751,8 +754,8 @@ public class ZMQ
             //  first pass have taken negligible time). We also compute the time
             //  when the polling should time out.
             if (firstPass) {
-                now = Clock.nowMS();
-                end = now + timeout;
+                now = Clock.nowUS();
+                end = now + TimeUnit.MILLISECONDS.toMicros(timeout);
                 if (now == end) {
                     break;
                 }
@@ -761,7 +764,7 @@ public class ZMQ
             }
 
             //  Find out whether timeout have expired.
-            now = Clock.nowMS();
+            now = Clock.nowUS();
             if (now >= end) {
                 break;
             }
--- a/src/main/java/zmq/util/Clock.java
+++ b/src/main/java/zmq/util/Clock.java
@@ -14,10 +14,20 @@ public class Clock
     {
     }
 
-    //  High precision timestamp.
+    /**
+     * High precision timestamp in microseconds.
+     */
     public static long nowUS()
     {
         return TimeUnit.MICROSECONDS.convert(System.nanoTime(), TimeUnit.NANOSECONDS);
+    }
+
+    /**
+     * High precision timestamp in nanoseconds.
+     */
+    public static long nowNS()
+    {
+        return System.nanoTime();
     }
 
     //  Low precision timestamp. In tight loops generating it can be


daveyarwood:
Hmm, the Travis tests failed for some non-obvious reason judging by the logs -- seems like maybe it was a problem on Travis' end. I've restarted the builds.

daveyarwood:
@pepijnve The code looks good to me -- I'll be especially happy with it if the tests pass.

I wonder if it would be possible to add a test case that demonstrates the problem (i.e. fails in the absence of your change, passes with it)? That would be great for guarding against regressions in the future. I can understand if it's hard to do that in this case, though, since it seems to be dependent upon the system clock time.

pepijnve:
The only way to test this that I can come up with is mocking the system clock. But that would mean turning `zmq.util.Clock` into an interface and allowing it to be injected. I'm happy to make that change if that's what you want, but it's not something I would personally be inclined to do.

pepijnve:
@daveyarwood I figured out what's wrong. `Clock.nowUS` is microseconds; not nanoseconds. I was wondering why it said `U` 😂 I'll get that fixed right away.

pepijnve:
The test that was hanging on Travis was failing locally for me as well. Seems like there's something in the poll logic that I'm missing. I've pushed a second attempt that consists of the minimal change required to fix things compared to the current code. Tests pass locally now.

fredoboulo:
Hi @pepijnve, the code looks good to me as well. As @daveyarwood I would prefer some tests to avoid regression but as long as behaviour is improved, that's enough for me.

fredoboulo:
I'm just thinking out loud, but would you consider creating a method in clock called nowNS returning nano time, as it may avoid one conversion when polling?

pepijnve:
Absolutely. I just went for smallest possible scope of changes in my 2nd attempt. I’ll push a second change that uses ns in a minute.

fredoboulo:
My bad, it does not avoid a conversion... Anyhow, I'm happy with or without your second commit :)

pepijnve:
It does avoid a hidden conversion in Clock. Every call to `nowUS` does a nano- to microsecond conversion. 

FYI, I had a look at the corresponding libzmq code in clock.cpp. That also makes use of monotonically increasing clocks. `System#currentTimeMillis` doesn't provide that guarantee.

The same problem exists in `SocketBase#send`, `SocketBase#receive` and in the timer related code in PollerBase. Would you like me to fix those as part of this PR as well so that all the code dealing with timeouts is consistent?

fredoboulo:
Who am I to disagree ? :)

So 

```
    public static long nowMS()
    {
        return TimeUnit.MILLISECONDS.convert(System.nanoTime(), TimeUnit.NANOSECONDS);
    }

```

would do the trick?

trevorbernard:
I would create an issue first describing what the problem is. Then depending on how you've describe the problem(s), do the simplest things to solve that problem.

pepijnve:
@fredoboulo indeed, I think this can all be reduced down to that one-method change in `Clock`

@trevorbernard will do

pepijnve:
> I'm just wondering about performance.

There's a statement in the code that `nowMs` is '10 to 100 times faster' than `nowUS`. As far as I can tell that was copy/pasted directly from the C++ code; not sure what the relative difference between `nanoTime` and `currentTimeMillis` is or how significant that would be in the context of polling loops.

You could argue that the performance difference is not relevant since the potentially faster variant (`currentTimeMillis`) is not reliable.

fredoboulo:
Hi @pepijnve 

I tried to make some comparisons yesterday:

```
package zmq.util;

import org.junit.Test;

public class ClockTest
{
    @Test
    public void testClocknowMS()
    {
        testclock(new Runnable()
        {

            @Override
            public void run()
            {
                TimeUnit.MILLISECONDS.convert(System.nanoTime(), TimeUnit.NANOSECONDS);
            }
        }, "Clock.nowMS()");
    }

    @Test
    public void testClocknowNS()
    {
        testclock(new Runnable()
        {

            @Override
            public void run()
            {
                System.nanoTime();
            }
        }, "Clock.nowNS()");
    }

    @Test
    public void testSystemcurrentTimeMillis()
    {
        testclock(new Runnable()
        {

            @Override
            public void run()
            {
                System.currentTimeMillis();
            }
        }, "System.currentTimeMillis()");
    }

    private void testclock(Runnable run, String desc)
    {
        final int iterations = 10000000;

        // warmup
        for (int idx = 0; idx < iterations; ++idx) {
            run.run();
        }

        final long start = System.nanoTime();

        for (int idx = 0; idx < iterations; ++idx) {
            run.run();
        }

        final long end = System.nanoTime();
        final double time = (end - start) / (double) iterations;
        System.out.println(desc + " takes " + time + " nanoseconds");
    }
}

```

Results are:

```
System.currentTimeMillis() takes 23.2335018 nanoseconds
Clock.nowMS() takes 25.2765916 nanoseconds
Clock.nowNS() takes 23.860775 nanoseconds

``` 

It's a bit rough as results are not that stable, but I do not see a huge performance gap. Would someone else please have a look as well?

pepijnve:
One of the replies at https://stackoverflow.com/questions/19052316/why-is-system-nanotime-way-slower-in-performance-than-system-currenttimemill hints that the performance difference is specific to Windows.

I found a copy of the Sun/Oracle blog post in the internet archive (https://web.archive.org/web/20160308031939/https://blogs.oracle.com/dholmes/entry/inside_the_hotspot_vm_clocks). That states that, on Windows, `currentTimeMillis` consists of a global variable read; `nanoTime` uses `QueryPerformanceCounter` (QPC). How QPC is actually implemented is Windows version and hardware dependent. See https://msdn.microsoft.com/en-us/library/windows/desktop/dn553408(v=vs.85).aspx

So in summary, the actual performance cost is JVM, hardware and operating system dependent. `nanoTime` can be slower than `currentTimeMillis`.

fredoboulo:
Thanks for the pointers. Still, it would be nice to have actual measurements. I only have Ubuntu so the numbers that I provided are only relevant on my machine Ubuntu 16.04 with jdk8, and TSC clock source.

pepijnve:
Agreed, actual measurements would be interesting. I’m afraid I don’t have any old hardware or Windows versions at my disposal to test this myself.

daveyarwood:
We do have Travis at our disposal. We are [currently running tests with a couple of different JDKs](https://github.com/zeromq/jeromq/blob/master/.travis.yml), and we could certainly add more.

That doesn't cover our bases with non-Linux OSes, but maybe it's a start?

trevorbernard:
I'm going to merge this pull request as it satisfies the C4 process.

Issue #520, opened 2018-02-07
IO thread loops infinitely on out event and blocks application with 100% CPU usage [was renamed]


smattheis:
**UPDATE: All assumptions in this comment turned out to be wrong. The true problem is described below in comment https://github.com/zeromq/jeromq/issues/520#issuecomment-364570301**

I'm using the latest version 0.4.3 of JeroMQ and my application stops after some time because messages are not sent over a PUSH socket while one of two JeroMQ I/O threads is at 100% CPU usage. The thread remains inside epollWait, see http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8u40-b25/sun/nio/ch/EPollArrayWrapper.java#269

I've seen issue https://github.com/zeromq/jeromq/issues/506 and think it's very likely related. I don't know about the defined behavior of epollWait and how it's supposed to be used in JeroMQ, but I see in my application that the timeout argument is always zero whenever it hangs. As I thought it should be some positive timeout value (*) I replaced `long timeout = executeTimers();` with `long timeout = Math.max(10, executeTimers());` but it doesn't solve the problem and the io thread still hangs forever inside epollWait. Since the thread hangs inside epollWait the selector recreate logic is not executed. Do you have any idea what's wrong?

(*) I saw that JeroMQ implements a similar logic to workaround the JDK bug as netty; however, in netty epollWait is called with a timeout of 10.

In my application, I execute a test that uses four PULL sockets and various PUSH sockets. The test is started multiple times (because I do a bunch of simulations) in the same Java process but each time with a new context. After each test, sockets are closed and the context is closed gracefully.

fredoboulo:
Hi, could you provide some tests for us to get a better understanding of your situation? Some information about your OS, java version , ... would be useful as well. Does it occur when loosing network, hibernating computer, ...?

smattheis:
In my test the error occurs with 90% probability in a 5 min execution time such that I can actually reproduce the problem quite reliably. However, I can't share that code. I tried to reproduce the problem in an isolated test but without success. Nevertheless, I will keep on trying to reproduce it.

[...]

PR #522, opened 2018-02-09
Bugfix: prevent corrupted buffer of encoder base on zero-copy encode

smattheis:
This PR is an attempt to fix a problem of potentially corrupted byte buffers that can cause io threads to loop on the same out event infinitely and, consequently, to block/deadlock applications. For details, see issue https://github.com/zeromq/jeromq/issues/520

### DIFF
--- a/src/main/java/zmq/io/StreamEngine.java
+++ b/src/main/java/zmq/io/StreamEngine.java
@@ -577,6 +577,7 @@ public class StreamEngine implements IEngine, IPollEvents
                 buf.flip();
             }
         }
+        assert (outsize == outpos.get().remaining());
 
         //  If there are any data to write in write buffer, write as much as
         //  possible to the socket. Note that amount of data to write can be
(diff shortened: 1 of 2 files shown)


smattheis:
It seems that the bugfix breaks the `V1EncoderTest` and `V2EncoderTest`since both tests initialize the encoder with a bufferSize of 64 while the bugfix adds a condition for zero-copy encoding to be a minimum size of OUT_BATCH_SIZE. At the same time the bugfix induces a dependency of the encoder on the Config which may break some architectural design decisions. I think this requires a decision of the maintainers of JeroMQ on how to approach that.
Further, please note that I have not checked if a similar bug may occur at other points in the code. I don't know the library so well to sufficiently analyze that or even to really trust own analysis.

smattheis:
The original issue is resolved by PR https://github.com/zeromq/jeromq/pull/525. I'm closing this PR respectively.

PR #525, opened 2018-02-26
Fix/520 encoder flip

fredoboulo:


fredoboulo:
#520 reported a bug in the encoder. While I'm not sure if it will totally solve the situation, it actually pinpointed an edge case that is solved by this PR.

Feel free to propose any other naming for methods!

PR #527, opened 2018-02-27
Fix #524 : V1 and V2 protocol downgrades handle received data in handshake buffer

### LINKED ISSUE #524: testProtocolVersion1short or testProtocolVersion2short fail indeterministically
I sometimes get a test failure where testProtocolVersion1short or testProtocolVersion2short fails. The reason of failure is timeout. Rising the timeout value does not help. testProtocolVersion1long and testProtocolVersion2long passes. Could you give me some pointers what can go wrong here?

fredoboulo:


Boskovits:
I'm now quite sure, that the other test failures are unrelated. This fix is fine, please merge.

PR #539, opened 2018-04-17
Heartbeat proposal (fixes #547)

### LINKED ISSUE #547: Implement heartbeating as of https://rfc.zeromq.org/spec:37/ZMTP
Current implementation does not provide heartbeating between sockets.
This would be nice-to-have so that it is compatible with other ZMQ implementations (as libzmq ,for instance).

fredoboulo:
I propose embedded heartbeating, should be compatible with latest libzmq on that particular regard, even if I did not test that compatibility part.

Still have some concerns about methods naming though, if anybody may give inputs it would be nice.

### DIFF
--- a/src/main/java/org/zeromq/ZMQ.java
+++ b/src/main/java/org/zeromq/ZMQ.java
@@ -724,6 +724,17 @@ public class ZMQ
         }
 
         /**
+         * The ZMQ_HEARTBEAT_CONTEXT option shall set the ping context
+         * of the peer for ZMTP heartbeats.
+         * If this option is set, every ping message sent for heartbeat will contain this context.
+         * @return the context to be sent with ping messages. Empty array by default.
+         */
+        public byte[] getHeartbeatContext()
+        {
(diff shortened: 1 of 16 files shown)


bluca:
@fredoboulo  @trevorbernard note that the heartbeat changed slightly in the RFC recently (ZMTP 3.1 is still in DRAFT state) - precisely, the ping context was restricted to 16 bytes max, anything over will be simply truncated. This is to prevent abuse, given the heartbeat runs in the I/O thread with no application control. Also there's really no need for more than that. See: https://github.com/zeromq/rfc/pull/130

Also note that the context is optional, and can not be present - libzmq does not set one for example. The TTL wording was also clarified (but not changed) in https://github.com/zeromq/rfc/pull/131

fredoboulo:
Thanks for the heads-up @bluca !

Any plans for setting this ping context in libzmq? There's now a simple draft option to set it statically in Jeromq, I'm not very keen on letting application build it dynamically within IO thread.

bluca:
To be honest, I'm not quite sure what's the utility of the context. Unless someone asks for it, I don't think I'll add an API to let users set it. But libzmq does now support it at the protocol level and will pong it back if sent.

PR #549, opened 2018-05-05
First cut of "getting started" documentation

### LINKED ISSUE #541: A FAQ is missing
A simple FAQ would be helpful.

When someone starts on jeromq, there is some simple questions what will help peoples identify pitfall.

There is a least three wrappers: zmq.Ctx, ZMQ.Context, ZContext. Which one to use ?
There is two pollers: ZPoller and ZMQ.Poller, which is the one returned by ZContext. Again, which one to use ?

A sample to use Curve can be found at https://gist.github.com/trevorbernard/6e3a8af0092cdced0f8b3e757a5b6b16, but it's not in src/test/java/guide. It could be part of the FAQ.

There is a lot of defined exceptions, that overlap and are declared in different ways:
```
org/zeromq/ZMQException.java:public class ZMQException extends RuntimeException
zmq/ZError.java:    public static class CtxTerminatedException extends RuntimeException
zmq/ZError.java:    public static class InstantiationException extends RuntimeException
zmq/ZError.java:    public static class IOException extends RuntimeException
```
Is there any guide about that ?

What about the package org.zeromq versus zmq. Why two root packages ? Is there any meaning about that ?

daveyarwood:
Closes #541.

I don't claim to be an authority on JeroMQ by any means, but I've documented what I know. Hopefully this addresses the specific questions raised in #541 by @fbacchella, and can serve as a good starting point for the project to have nice, user-friendly documentation.

NB: I considered putting this information in the Javadocs, but it occurred to me that Javadocs are organized by packages and classes, which doesn't give us a good place to document general ideas about ZeroMQ and JeroMQ. For this reason, I think it's better to have a canonical "docs" folder where we can organize things in a more intuitive way.

Feedback welcome!

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,13 +17,13 @@ Pure Java implementation of libzmq (http://zeromq.org).
 * Securities
  * [PLAIN](http://rfc.zeromq.org/spec:24).
  * [CURVE](http://rfc.zeromq.org/spec:25).
- 
-* Not too bad performance compared to zeromq.
+
+* Performance that's not too bad, compared to native libzmq.
  * 4.5M messages (100B) per sec.
  * [Performance](https://github.com/zeromq/jeromq/wiki/Performance).
 * Exactly same developer experience with zeromq and jzmq.
 
-## Not supported Features
+## Unsupported
 
 * ipc:// protocol with zeromq. Java doesn't support UNIX domain socket.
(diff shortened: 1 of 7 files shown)


fredoboulo:
Looks great! Thanks @daveyarwood !

PR #552, opened 2018-05-06
Problem: HEARTBEAT command breaks REQ connection

### LINKED ISSUE #551: Cannot use heartbeat with REQ
This is the same issue as https://github.com/zeromq/libzmq/issues/3060

fredoboulo:
Solution: ignore command messages in the REQ session to avoid
disrupting the state machine.
Commands are handled by the engine before handing off to the session.

Fixes #551

### DIFF
--- a/src/test/java/zmq/HeartbeatsTest.java
+++ b/src/test/java/zmq/HeartbeatsTest.java
@@ -16,6 +16,51 @@ import zmq.util.Utils;
 
 public class HeartbeatsTest
 {
+    // this checks for heartbeat in REQ socket.
+    @Test
+    public void testHeartbeatReq() throws IOException
+    {
+        final int heartbeatInterval = 100;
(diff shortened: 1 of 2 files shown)


src/test/java/zmq/HeartbeatsTest.java
```diff
@@ -16,6 +16,50 @@
...
+        SocketBase rep = ctx.createSocket(ZMQ.ZMQ_REP);
+        assertThat(rep, notNullValue());
+        rc = rep.bind(addr);
+        assertThat(rc, is(true));
+
+        final long start = System.currentTimeMillis();
+        do {
+            // request
+            int ret = ZMQ.send(req, "hello", 0);
+            assertThat(ret, is(5));
```
daveyarwood: It isn't clear what this value 5 means. It would be good to see a named constant, so that it is self-documenting.

PR #556, opened 2018-05-28
Problem: using int flags to indicate socket type is error-prone

esahekmat:
Solution: create a SocketType enum and use it instead of int flags
also mark int flag and their methods with Deprecated annotation


daveyarwood:
+1, this looks great! Thanks for the PR! :+1:

PR #557, opened 2018-05-29
Problem: cannot use Java 8 lambda syntax to create IAttachedRunnable and IDetachedRunnable

esahekmat:
Solution: As Java 7 is obsolete long time ago, I think its time to move on and upgrade java version to use Java 8 new features and syntax.


src/test/java/org/zeromq/TestZThread.java
```diff
@@ -1,40 +1,30 @@
...
-                assertThat(push, notNullValue());
-
-                ctx.close();
-                stopped.countDown();
-            }
-        };
-
-        ZThread.start(detached);
+        ZThread.start((args) -> {
+            ZContext ctx = new ZContext();
```
daveyarwood: This would be a good place to use a try-with-resources block:

```java
try (ZContext ctx = new ZContext()) {
    Socket push = ctx.createSocket(SocketType.PUSH);
    assertThat(push, notNullValue());
}
```

daveyarwood:
Looks good, BTW :+1:

trevorbernard:
One thing to consider is this will likely break codebases in the wild. I'm not against upgrading to lowest supported java to java 8 but it's worth noting and talking about. I'm for changing the lowest supported java version.

esahekmat:
I think it's worth, I have tried to show some benefit with FuntionalInterface, but there is a lot of things there waiting to change, I think public API of JeroMQ should change a lot to attract the more modern developer, And one of these change is providing lambda API. I agree with you about breaking clients codebase, because of that I try a lot to still keep old API but upgrading lowest supported Java version has no other way. Maybe it's better to reflect this change with major version change, Maybe It's time for JeroMQ to go to version 5.0, clients who upgrade their JeroMQ to 5.0 should be aware to also upgrade their java version

trevorbernard:
There have been a number of people who've tried to create an Idiomatic java interface to JeroMQ, myself and @fredoboulo included. https://github.com/fredoboulo/zeromq-java-api https://github.com/zeromq/jzmq-api.

esahekmat:
I have looked them, Nice. specially jzmq-api looks good. but why they are inactive? would you share your experience in that case? is it hard to synchronize this project with the underlying library?
And also another question, It seems to me, JeroMQ API wants to look like zeroMQ C library. It tries hard to keep it same as C library, I think maybe JeroMQ developers want when someone read zguide, Java Examples and C Examples does not look different, am I right? I want to know if I'm right, This decision is fixed and inflexible?

esahekmat:
Thanks @trevorbernard, your answers clarified my thoughts. As you asked for dialog around Java 8, I think others should join discussion, but as an advocate of upgrading I want to notice another benefit of upgrading, I have planned to develop stream API for sockets to send and receive messages, we can use Consumer Interface for income messages and Producer Interface for outcome messages.

fredoboulo:
Looks good to me :)

I have no objection about switching to java8, only 2 remarks about it: 

- it would be nice to hear from Android developers so no-one is locked in their dev process.
- if we initiate the rule of switching versions of java when they become obsolete, we will have to talk about it again [next January](http://www.oracle.com/technetwork/java/eol-135779.html)

fredoboulo:
I cannot talk about jzmq-api, the other project is a try to create a common abstraction so devs can switch between jzmq and jeromq without problems. 
There is currently no common classes now, only classes that happen to have the same name and not-exactly-quite-same signatures.

As for sticking with libzmq, I would prefer it that way. It makes it easier to upgrade as Trevor stated. That does not mean there will be no small deviation, of course.

trevorbernard:
I spoke with some people who use JeroMQ and Android, and they didn't see anything wrong with it.

PR #559, opened 2018-06-09
problem: cannot treat incoming messages as an infinite stream

esahekmat:
solution: expose stream API in ZMQ.Socket class


fredoboulo:
I would consider annotate the new methods with @Draft and with a comment "This API is in DRAFT state and is subject to change at ANY time until declared stable."

That would allow to continue working on it until you are satisfied, with a fair warning to users.

Would that be interesting for you, @esahekmat ?

esahekmat:
You're right, I submit another PR for it

PR #560, opened 2018-06-11
Problem: Cannot chain call on ZMsg methods

esahekmat:
Solution: Use builder pattern where possible

### DIFF
--- a/src/main/java/org/zeromq/ZMsg.java
+++ b/src/main/java/org/zeromq/ZMsg.java
@@ -83,9 +83,10 @@ public class ZMsg implements Iterable<ZFrame>, Deque<ZFrame>
      * @param str
      *              String to add to list
      */
-    public void addString(String str)
+    public ZMsg addString(String str)
     {
         frames.add(new ZFrame(str));
+        return this;
     }
 
     /**
@@ -113,12 +114,13 @@ public class ZMsg implements Iterable<ZFrame>, Deque<ZFrame>
      * Message takes ownership of frame, will destroy it when message is sent.
      * @param frame
      */
-    public void wrap(ZFrame frame)
+    public ZMsg wrap(ZFrame frame)
     {
         if (frame != null) {
             push(new ZFrame(""));
             push(frame);
         }
+        return this;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/test/java/org/zeromq/ZMsgTest.java
```diff
@@ -123,20 +122,18 @@ public void testHashcode()
...
     {
-        ZMsg msg = new ZMsg();
-
-        msg.add(new byte[0]);
-        msg.add(new byte[] { (byte) 0xAA });
-        msg.dump();
+        ZMsg msg = new ZMsg()
+                .add(new byte[0])
+                .add(new byte[] { (byte) 0xAA })
+                .dump();
```
daveyarwood: Generally I'm in favor of the builder pattern, but it feels a little weird to me to include `.dump()` in the chain here, since that is an action that you perform using the ZMsg you've built. I would write it like this:

```java
ZMsg msg = new ZMsg()
        .add(new byte[0])
        .add(new byte[] { (byte) 0xAA });

msg.dump();

StringBuilder out = new StringBuilder();
msg.dump(out);
```

esahekmat: I think dump method usually used for log or debug purpose, I consider it something like peek() method in java8 streams, where you want to do some other things in the middle of work without any side-effect. Maybe I'm wrong, however I don't insist on builder pattern for dump method.

daveyarwood: I think it's probably valuable to have the builder pattern available for `dump` (it can't hurt, right?), but perhaps we could change the usage example here in the tests to reflect recommended usage.

esahekmat: I Agree

src/main/java/org/zeromq/ZMsg.java
```diff
@@ -396,55 +398,65 @@ public void dump(Appendable out)
...
     }
 
-    public void push(byte[] data)
+    public ZMsg push(byte[] data)
     {
         push(new ZFrame(data));
+        return this;
     }
 
-    public boolean add(String stringValue)
```
fredoboulo: Seeing it a bit late, but this breaks API, as the return value is not consistent anymore.
We may consider keeping the signature and adding a new method for the chaining call. I just do not know which name to use: and(), plus(), ...

esahekmat: I saw this boolean is always true, so I thought it's redundant and I can remove it. for the name of the new method I recommend addAnd() because two methods with the names add() and plus() in one class, could be a cause of misleading

PR #562, opened 2018-06-12
Use Java 8 default implementation of interfaces

fredoboulo:


fredoboulo:
@daveyarwood this sporadically failing test is annoying, I may have a solution to implement last endpoint with wildcard, that would also remove the usage of Utils.findOpenPort()

daveyarwood:
@fredoboulo I'm on-board with removing `Utils.findOpenPort()` -- it's not the most reliable method in the world!

esahekmat:
This PR make the code more clear and legible, Thanks :+1:

PR #567, opened 2018-06-28
Problem: various public APIs were removed or modified

fredoboulo:
Solution: Restore previously published APIs

### DIFF
--- a/src/main/java/org/zeromq/SocketType.java
+++ b/src/main/java/org/zeromq/SocketType.java
@@ -1,13 +1,13 @@
 package org.zeromq;
 
-import zmq.ZMQ;
-
 /**
  * Socket Type enumeration
  *
  * @author Isa Hekmatizadeh
  */
-public enum SocketType {
+@SuppressWarnings("deprecation")
+public enum SocketType
+{
     /**
      * Flag to specify a exclusive pair of sockets.
      * <p>
@@ -33,7 +33,7 @@ public enum SocketType {
      * and do not implement functionality such as auto-reconnection.
      * PAIR sockets are considered experimental and may have other missing or broken aspects.</strong>
      */
-    PAIR(ZMQ.ZMQ_PAIR),
+    PAIR(ZMQ.PAIR),
 
     /**
      * Flag to specify a PUB socket, receiving side must be a SUB or XSUB.
@@ -59,7 +59,7 @@ public enum SocketType {
      * <tr><td>Action in mute state</td><td>Drop</td></tr>
      * </table>
      */
-    PUB(ZMQ.ZMQ_PUB),
+    PUB(ZMQ.PUB),
 
     /**
      * Flag to specify the receiving part of the PUB or XPUB socket.
@@ -80,7 +80,7 @@ public enum SocketType {
      * <tr><td>Outgoing routing strategy</td><td>N/A</td></tr>
      * </table>
      */
-    SUB(ZMQ.ZMQ_SUB),
+    SUB(ZMQ.SUB),
 
     /**
      * Flag to specify a REQ socket, receiving side must be a REP or ROUTER.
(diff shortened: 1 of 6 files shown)


src/main/java/org/zeromq/ZMQ.java
```diff
@@ -757,9 +757,22 @@ public void close()
          * @param type    the socket type.
          */
         protected Socket(Context context, SocketType type)
+        {
+            this(context, type.type);
```
esahekmat: I Think to reference a Deprecated method is not a good idea, I prefer reverse of this, deprecated method call newly created method as in this situation removing deprecated method has no side effect

src/main/java/org/zeromq/ZMQ.java
```diff
@@ -797,9 +810,21 @@ public void close()
          *
          * @return the socket type.
          */
-        public SocketType getType()
```
esahekmat: I broke this API and I'm sorry, thank you for correcting this. :+1:

src/main/java/org/zeromq/ZMsg.java
```diff
@@ -447,15 +447,37 @@ public ZMsg push(byte[] data)
...
+        return add(new ZFrame(data));
+    }
+
+    /**
+     * Adds a string as a new frame in the message.
+     *
+     * @param stringValue the value to add
+     * @return this
+     */
+    public ZMsg append(String stringValue)
```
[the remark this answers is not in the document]
fredoboulo: Yes the discussion was pending but stopped progress. this PR is a way to reactivate it with code proposal.
After checking, there is already an append(ZMsg) in the class. With a third one, for example addAnd, we would end up with #add, #append #addAnd. This is not very consistent, hence my proposal. I would even make a more bold move, leaving add methods unchanged, and name the ones with chaining calls 'append'. That would be simpler to understand. What would you think of that?

src/main/java/org/zeromq/SocketType.java
```diff
@@ -33,7 +33,7 @@
      * and do not implement functionality such as auto-reconnection.
      * PAIR sockets are considered experimental and may have other missing or broken aspects.</strong>
      */
-    PAIR(ZMQ.ZMQ_PAIR),
```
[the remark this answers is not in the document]
fredoboulo: There is still a goal to be consistent with jzmq. The only common API between them if org.zeromq, it would make things simpler.

src/main/java/org/zeromq/ZMQ.java
fredoboulo: It's internal change, this one is faster, as the int type is already available. If you swap constructors callings, code will have to iterate each time to find out the proper enum.

src/main/java/org/zeromq/ZMQ.java
fredoboulo: Neh, no problem, no need to say sorry :)

src/main/java/org/zeromq/ZMsg.java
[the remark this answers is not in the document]
esahekmat: Your right, append(ZMsg) already exist, so it's better to have two method name, add and append.

src/main/java/org/zeromq/ZMQ.java
esahekmat: I understand, but creating a socket is not in the critical path.

fredoboulo: Of course, but for the sake of compatibility with previous releases, this constructor will stay in place for several versions. Doing it that way is a bit faster, so I propose to stay pragmatic and use it for the time being. When the time will come to remove the deprecated constructor, we can recode the other.

src/main/java/org/zeromq/SocketType.java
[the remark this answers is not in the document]
fredoboulo: I would even propose to use zmq.socket.Sockets but I sincerely hope we will find a way to create a common API with jzmq. Then it would be harder if there are too many references to zmq packages.

fredoboulo:
This PR is stalled for quite some time without info about why. Is there anything missing regarding C4 process or any other topic?

PR #581, opened 2018-07-25
Stops polling when selector is asynchronously closed.

### LINKED ISSUE #580: ClosedSelectorException when polling indefinitely and poller is closed before context is closed
By following these steps I am able to reliably crash JeroMQ with a ClosedSelectorException:

1. Create a ZContext and a ZMQ.Poller in a try-with-resource block.
2. Create a socket and add it to the poller with POLLIN.
3. Create a thread that uses the poller to poll(-1).
4. Wait for the thread to start.
5. Return on the main thread so the try-with-resource triggers close() calls.

Here's the test (the sleep is there just to be extra sure it crashes; it crashed reliably without the sleep on my machine):
```java
import org.zeromq.ZContext;
import org.zeromq.ZMQ;

public class Test {

	public static void main (String[] args) throws Exception {
		try (final ZContext context = new ZContext(1);
		     final ZMQ.Poller topPoll = context.createPoller(1)) {
			ZMQ.Socket pairSocket = context.createSocket(ZMQ.PAIR);

			topPoll.register(pairSocket, ZMQ.Poller.POLLIN);

			new Thread(() -> {
				try {
					topPoll.poll(-1);
				} catch (ZMQException e) {
					e.printStackTrace();
				}
			}).start();
			// May not be necessary
			Thread.sleep(100);
		}
	}
}
```

I get the following exception:
```
Exception in thread "Thread-0" java.nio.channels.ClosedSelectorException
	at sun.nio.ch.SelectorImpl.keys(SelectorImpl.java:68)
	at zmq.ZMQ.poll(ZMQ.java:722)
	at org.zeromq.ZMQ$Poller.poll(ZMQ.java:3210)
	at Test.lambda$main$0(Test.java:16)
	at java.lang.Thread.run(Thread.java:748)
```

The workaround is to never close Poller's, but to let the ZContext clean them up as it closes. Then, the fix from this pull request https://github.com/zeromq/jeromq/pull/390 triggers and handles the exception for us.

fredoboulo:
This may happen when the ZContext is closed in a separate thread.

Fixes #580


fredoboulo:
@daveyarwood the opposite could be done as well: catch the ClosedSelectorException at the same place as the ClosedChannelException and return -1.

ClosedSelectorException is kinda general problem: no polling can be performed on sockets. ClosedChannel is per channel so per socket.

PR #583, opened 2018-07-30
Race condition with received message causes ZMQ_CONNECT_RID to be assigned to wrong socket

mvilim:
Due to a race condition between receiving connections on the io thread and the call to socket.connect(), ZMQ_CONNECT_RID can be assigned to the wrong peer socket.

I recently fixed the same bug in the [libzmq](https://github.com/zeromq/libzmq/issues/3191) implementation. You can see more details there.


mvilim:
It appears the test that is failing (TermEndpointIpcTest) has been intermittently failing on the master branch, so I believe that is flaky test and unrelated to my commit (it is using a bind to random port and then release as a way to find a free port which apparently is not reliable).

PR #584, opened 2018-07-30
Fix flaky test when trying to check behaviour of disconnect and unbind

fredoboulo:


fredoboulo:
yes, but the error occurs in ipc:// protocol when binding to a non-random port (not a wildcard).

I will change the while condition to the following

```
...
        } while (!rc && push.errno() == ZError.EADDRINUSE);
...
```

PR #590, opened 2018-08-23
Selector can be created and closed from ZContext

fredoboulo:


daveyarwood:
This code looks good, but just out of curiosity, what is the motivation behind managing selectors directly? To me, they seem like implementation details of pollers that users shouldn't ever have to touch.

fredoboulo:
The main motivation is that ZContext allows to create a selector but does not provide any way to close it and remove it from internal list. While I was at it, I added a way to take advantage of this new API for ZStar.

Unless I misunderstood the question (quite late here), this is the opposite of managing them directly.

fredoboulo:
Right. Should we deprecate ZContext#createSelector() instead? I would still keep the closeSelector method, without it any caller of the former method will release resources of the selector when closing the context. It's too late to be optimal.
Mmmm... this branch was stale for one year in my repo, maybe I should have thought about it a bit more before trying to flush it.

daveyarwood:
Yeah, given that we already have `createSelector`, I think it's reasonable to add `closeSelector`, but also deprecate both of those with an explanation that the caller shouldn't ever have to manage selectors directly, as the context does this for you under the hood.

fredoboulo:
We still keep access of the equivalent methods in ZMQ.Context, they are mandatory.

src/main/java/org/zeromq/ZContext.java
```diff
@@ -170,9 +182,44 @@ public void destroySocket(Socket s)
...
+    }
+
+    /**
+     * Closes a selector.
+     * This is a DRAFT method, and may change without notice.
+     *
+     * @param selector the selector to close. It needs to have been created by {@link #createSelector()}.
+     * @deprecated {@link #createSelector()} was exposed by mistake. while waiting for the API to disappear, this method is provided to allow releasing resources.
+     */
+    @Deprecated
```
fredoboulo: It feels weird to release a new method that is deprecated straight away. We might want to not deprecate it and only mark it as draft. Thoughts?

daveyarwood: I'm not sure I understand why we need to add this method, if we don't intend for anyone to use it! :)   IMHO, we should just deprecate  `createSelector`.

fredoboulo: I start to be lost here by what you want to achieve. Better to make this PR then you can recode what you wish is necessary.

daveyarwood: So, the problem here, in my mind (and which you captured in #592), is that we expose a method `createSelector`, which is really an implementation detail, since selectors should be managed internally by the context.

A separate issue is that there is no `closeSelector` method, but that's only a problem because `createSelector` exists. Ideally, no one should call `createSelector` because you shouldn't be managing selectors yourself -- you should leave it up to the context.

PR #599, opened 2018-09-14
Add LGTM code quality badges

xcorail:
Hi there!

I thought you might be interested in adding these [code quality badges](https://lgtm.com/projects/g/zeromq/jeromq/ci/#badges) to your project. They will indicate a good code quality to potential users and contributors.
To get an idea of the analyses reflected by these grades, check the [alerts discovered by LGTM](https://lgtm.com/projects/g/zeromq/jeromq).

N.B.: I am on the team behind LGTM.com, I'd appreciate your feedback on this initiative, whether you're interested or not, if you find time to drop me a line. Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,8 @@ Pure Java implementation of libzmq (http://zeromq.org).
 [![Coverage Status](https://coveralls.io/repos/github/zeromq/jeromq/badge.svg?branch=master)](https://coveralls.io/github/zeromq/jeromq?branch=master)
 [![Maven Central](https://img.shields.io/maven-central/v/org.zeromq/jeromq.svg)](https://maven-badges.herokuapp.com/maven-central/org.zeromq/jeromq)
 [![Javadocs](http://www.javadoc.io/badge/org.zeromq/jeromq.svg)](http://www.javadoc.io/doc/org.zeromq/jeromq)
+[![Code Quality: Java](https://img.shields.io/lgtm/grade/java/g/zeromq/jeromq.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/zeromq/jeromq/context:java)
+[![Total Alerts](https://img.shields.io/lgtm/alerts/g/zeromq/jeromq.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/zeromq/jeromq/alerts)
 
 ## Features
 


trevorbernard:
What problem does this solve?

xcorail:
Hi @trevorbernard 

The badge itself doesn't solve any problem.
However, you might be interested in fixing these [alerts](https://lgtm.com/projects/g/zeromq/jeromq/alerts/?mode=tree). 

Note that you don't need to accept this PR and put the badge to have access to LGTM. The alerts will always be there for you on a daily basis, or even immediately if you integrate LGTM in your PRs ([one-click and free](https://lgtm.com/projects/g/zeromq/jeromq/ci/)).

daveyarwood:
I vote that we close this PR without merging, in the interests of keeping the README from getting too cluttered.

LGTM looks cool, though, thanks for the PR anyway!

fredoboulo:
Désolé Xavier, je respecte le travail fourni par votre équipe et je mesure les alertes apportées sur le projet.
Néanmoins, et au-delà de l'effet immédiat, la plus-value me semble minimale pour le projet lui-même en l'état actuel.

Thanks to you, we will be able to fix (and remove) some lines, my personal opinion would be to ideally integrate these alerts even before submitting a PR.

About the false positives, I fail to see here and now how they could be solved indirectly from your PR, especially regarding Options class. 

You might want to take a look at the alert to StreamEngine, it should definitely not appear, an assert in our code is guaranteeing that, just few lines above.

Je rejoins Dave: votre projet me semble très intéressant! Je vous souhaite une bonne fortune!

xcorail:
Hi @fredoboulo 
>Désolé Xavier, je respecte le travail fourni par votre équipe et je mesure les alertes apportées sur le projet.
Néanmoins, et au-delà de l'effet immédiat, la plus-value me semble minimale pour le projet lui-même en l'état actuel.

Je comprends. 

> Thanks to you, we will be able to fix (and remove) some lines, my personal opinion would be to ideally integrate these alerts even before submitting a PR.

Thanks for this feedback. I strongly agree, this is a more interesting approach. Now I need to find more fingers to do that ... ;-)

> About the false positives, I fail to see here and now how they could be solved indirectly from your PR, especially regarding Options class.

Well, the open source community is a great help for us, by reporting FP when they see them, on our [forum](https://discuss.lgtm.com/), or by directly proposing a fix on https://github.com/Semmle/ql. BTW, the one you reported appears indeed to be a false positive, and I will raise it to my Java colleagues 👍 . 
But this is a very indirect link, I agree.

> Je rejoins Dave: votre projet me semble très intéressant! Je vous souhaite une bonne fortune!

Merci beaucoup! If you intend to fix some alerts now, know that a competition started yesterday, which is just about that. The first prize is a trip to San Francisco for Github Universe conference, and there is also a donation to WWF for each alert fixed: https://competitions.lgtm.com/ghu-2018 

Good luck to all jeromq team and thanks again for the valuable feedback.

xcorail:
@fredoboulo coming back to StreamEngine.java, I think it's a true positive: 
The assert checks
`assert (options.mechanism == Mechanisms.NULL)`
but the alert says that `mechanism` can be null (the real null, not `Mechanisms.NULL`), because there is a null guard later
`if (options.mechanism == null)`

So basically the alert says that if this later condition `options.mechanism == null` may be true, then it means that `options.mechanism` may be null. 
Of course, only you can tell, according to the code, if the null guard is useless because `options.mechanism` is never null, or it should actually be tested.

Would you please check before I report this as a FP to my Java team?
Cheers

fredoboulo:
Riiiiight...

Ok, you found a dead branch of the code then :) This (mechanism == null) should never happen.
That's for the part of jeromq.

For LGTM, the assert just above still assess that the variable cannot be null, as it asserts the value in a specific set of values.

So actually, your alert is about the useless test condition. Interesting!

PR #602, opened 2018-09-18
Fix #601 - Change to use eu.neilalexander/jnacl

### LINKED ISSUE #601: Problem: no need to maintain jnacl anymore since original author released to central

trevorbernard:


daveyarwood:
Seems reasonable to me, and the build is passing :+1:

PR #609, opened 2018-09-23
Problem: Arguments of public APIs are checked with asserts. Fixes #605

### LINKED ISSUE #605: Public APIs parameters are checked with asserts
Asserts can be disabled and should be used only for internal invariants.

fredoboulo:
Solution: unconditional checking of arguments of public methods and constructors for org.zeromq

### DIFF
--- a/src/main/java/org/zeromq/ZMonitor.java
+++ b/src/main/java/org/zeromq/ZMonitor.java
@@ -4,6 +4,7 @@ import java.io.Closeable;
 import java.io.IOException;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 import java.util.UUID;
 
 import org.zeromq.ZMQ.Socket;
@@ -44,7 +45,7 @@ public class ZMonitor implements Closeable
         private ZEvent(ZMsg msg)
         {
             assert (msg != null);
-            assert (msg.size() == 3 || msg.size() == 4);
+            assert (msg.size() == 3 || msg.size() == 4) : msg.size();
 
             type = Event.valueOf(msg.popString());
             code = Integer.valueOf(msg.popString());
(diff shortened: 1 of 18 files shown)


src/main/java/zmq/Mailbox.java
```diff
@@ -89,7 +89,7 @@ public Command recv(long timeout)
         //  Wait for signal from the command sender.
         boolean rc = signaler.waitEvent(timeout);
         if (!rc) {
-            assert (errno.get() == ZError.EAGAIN || errno.get() == ZError.EINTR);
+            assert (errno.get() == ZError.EAGAIN || errno.get() == ZError.EINTR) : errno.get();
```
daveyarwood: I suppose there might be some tiny possibility that `errno` is reset in another thread while this line is being executed, so perhaps we should read the value one time only? i.e.

```java
int errno = errno.get();
assert (errno == ZError.EAGAIN || errno == ZError.EINTR) : errno;
```

src/main/java/org/zeromq/ZMonitor.java
```diff
@@ -44,7 +45,7 @@
         private ZEvent(ZMsg msg)
         {
             assert (msg != null);
-            assert (msg.size() == 3 || msg.size() == 4);
+            assert (msg.size() == 3 || msg.size() == 4) : msg.size();
```
daveyarwood: Similar to me other comment, it might be safer if we read the size only once:

```java
int size = msg.size();
assert (size == 3 || size == 4) : size;
```

src/main/java/zmq/Mailbox.java
fredoboulo: I'm pretty confident that it cannot happen: Errno holds a ThreadLocal variable, its value is per-thread.

src/main/java/org/zeromq/ZMonitor.java
fredoboulo: msg is passed into the constructor and is not used otherwise. There is no possibility it can be modified at that point in time.

PR #614, opened 2018-10-19
A SelectorProvider

fbacchella:
I added a way to define custom SelectorProvider (https://docs.oracle.com/javase/8/docs/api/java/nio/channels/spi/SelectorProvider.html). It should allow to wrap socket with custom code, like flow compression or TLS, using for example https://github.com/marianobarrios/tls-channel.
Any feedback is welcome.

### DIFF
--- a/src/main/java/zmq/Options.java
+++ b/src/main/java/zmq/Options.java
@@ -7,6 +7,7 @@ import java.util.List;
 import zmq.io.coder.IDecoder;
 import zmq.io.coder.IEncoder;
 import zmq.io.mechanism.Mechanisms;
+import zmq.io.net.SelectorProviderChooser;
 import zmq.io.net.ipc.IpcAddress;
 import zmq.io.net.tcp.TcpAddress;
 import zmq.io.net.tcp.TcpAddress.TcpAddressMask;
@@ -163,6 +164,8 @@ public class Options
     // Set to <= 0 to disable this system.
     public MsgAllocator allocator;
 
+    public SelectorProviderChooser selectorChooser;
+
     public final Errno errno = new Errno();
 
     public Options()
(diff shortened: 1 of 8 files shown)


src/main/java/zmq/Options.java
```diff
@@ -519,6 +524,27 @@ else if (optval instanceof MsgAllocator) {
...
+            }
+            else if (optval instanceof Class) {
+                selectorChooser = chooser((Class<?>) optval);
+                return true;
+            }
+            else if (optval instanceof SelectorProviderChooser) {
+                selectorChooser = (SelectorProviderChooser) optval;
+                return true;
+            }
+            selectorChooser = (SelectorProviderChooser) optval;
```
fredoboulo:
```suggestion
            return false
```

src/test/java/zmq/io/net/SelectorProviderTest.java
```diff
@@ -0,0 +1,65 @@
+package zmq.io.net;
+
+import java.io.IOException;
+import java.nio.channels.spi.SelectorProvider;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.zeromq.SocketType;
```
fredoboulo: We could consider splitting the tests into 2 classes, one in zmq namespace, the other one in org.zeromq.

fredoboulo:
I really like this, I'm wondering if that would allow the tests of sad paths of the lib that are currently uncovered.

src/main/java/zmq/Options.java
fbacchella: Oups, my bad.

fbacchella:
Thank you ! Any hop to see a 0.4.4 soon ? The last release is almost one year old.

PR #618, opened 2018-10-28
Removes several printStackTrace dumped into stderr

fredoboulo:
This PR does not remove all of them, and I still get some doubts about some parts. I will comment the doubtful lines in the PR itself.

### DIFF
--- a/src/main/java/org/zeromq/ZStar.java
+++ b/src/main/java/org/zeromq/ZStar.java
@@ -10,6 +10,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.BiFunction;
 
+import org.zeromq.ZMQ.Error;
 import org.zeromq.ZMQ.Socket;
 import org.zeromq.ZThread.IAttachedRunnable;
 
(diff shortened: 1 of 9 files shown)


src/main/java/org/zeromq/ZStar.java
```diff
@@ -520,7 +520,6 @@ public void run(final Object[] train, final ZContext chef, final Socket mic)
                         }
                         catch (Exception e) {
                             // really ?
-                            e.printStackTrace();
```
daveyarwood: I'm curious about this one. Swallowing exceptions is usually bad news, especially unspecific exceptions. Probably better to remove the try/catch altogether and let exceptions happen, if this is a "this should never happen" scenario. Or we could wrap it in some kind of ZMQ exception wrapper class.

fredoboulo:
@daveyarwood If you are about to make a release, merging this PR can wait. Last-second changes? Mmmm better wait actually. I will close the PR and reopen it later.

daveyarwood:
@fredoboulo Sounds good :)  

I think @trevorbernard will be cutting a release soon.

trevorbernard:
Yeah, 0.4.3 is almost 1 year old.

PR #623, opened 2018-11-02
fixes #622

### LINKED ISSUE #622: Matching Batch-Sizes and Buffer-Sizes
User do not have the option to specify Config.IN_BATCH_SIZE nor Config.OUT_BATCH_SIZE. However, in case the user specifies large Options.rcvbuf or Options.sndbuf (in out case the payload of one message is ~16MB) it could be worthwhile to adjust Config.IN_BATCH_SIZE or  Config.OUT_BATCH_SIZE accordingly in order to reduce system calls.

markif:
Addresses issue #622

### DIFF
--- a/src/main/java/zmq/io/StreamEngine.java
+++ b/src/main/java/zmq/io/StreamEngine.java
@@ -255,8 +255,8 @@ public class StreamEngine implements IEngine, IPollEvents
         ioError = false;
 
         //  Make sure batch sizes match large buffer sizes
-        final int inBatchSize = options.rcvbuf > Config.IN_BATCH_SIZE.getValue() ? options.rcvbuf : Config.IN_BATCH_SIZE.getValue();
-        final int outBatchSize = options.sndbuf > Config.OUT_BATCH_SIZE.getValue() ? options.sndbuf : Config.OUT_BATCH_SIZE.getValue();
+        final int inBatchSize = Math.max(options.rcvbuf, Config.IN_BATCH_SIZE.getValue());
+        final int outBatchSize = Math.max(options.sndbuf, Config.OUT_BATCH_SIZE.getValue());
 
         if (options.rawSocket) {
             decoder = (IDecoder) instantiate(options.decoder, inBatchSize, options.maxMsgSize);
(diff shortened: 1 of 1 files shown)


src/main/java/zmq/io/StreamEngine.java
```diff
@@ -254,14 +254,18 @@ public void plug(IOThread ioThread, SessionBase session)
         handle = ioObject.addFd(fd);
         ioError = false;
 
+        //  Make sure batch sizes match large buffer sizes
+        final int inBatchSize = options.rcvbuf > Config.IN_BATCH_SIZE.getValue() ? options.rcvbuf : Config.IN_BATCH_SIZE.getValue();
+        final int outBatchSize = options.sndbuf > Config.OUT_BATCH_SIZE.getValue() ? options.sndbuf : Config.OUT_BATCH_SIZE.getValue();
```
daveyarwood: These could be simplified using Math.max:

```java
final int inBatchSize = Math.max(options.rcvbuf, Config.IN_BATCH_SIZE.getValue());
final int outBatchSize = Math.max(options.sndbuf, Config.OUT_BATCH_SIZE.getValue());
```

src/main/java/zmq/io/StreamEngine.java
```diff
@@ -469,13 +473,16 @@ public void outEvent()
             outpos.set(null);
             outsize = encoder.encode(outpos, 0);
 
-            while (outsize < Config.OUT_BATCH_SIZE.getValue()) {
+            //  Make sure batch sizes match large buffer sizes
+            final int outBatchSize = options.sndbuf > Config.OUT_BATCH_SIZE.getValue() ? options.sndbuf : Config.OUT_BATCH_SIZE.getValue();
```
daveyarwood: same deal here re: Math.max

src/main/java/zmq/io/StreamEngine.java
```diff
@@ -607,6 +614,10 @@ private boolean handshake()
         //  Position of the version field in the greeting.
         final int revisionPos = SIGNATURE_SIZE;
 
+        //  Make sure batch sizes match large buffer sizes
+        final int inBatchSize = options.rcvbuf > Config.IN_BATCH_SIZE.getValue() ? options.rcvbuf : Config.IN_BATCH_SIZE.getValue();
+        final int outBatchSize = options.sndbuf > Config.OUT_BATCH_SIZE.getValue() ? options.sndbuf : Config.OUT_BATCH_SIZE.getValue();
```
daveyarwood: and here as well

src/main/java/zmq/io/StreamEngine.java
markif: Seems reasonable.

markif:
Thanks for your comments and sorry for the delay (must have slipped through the cracks). I applied the corresponding changes. I neither know the difference/history between `options` and `Config`. `Config` seems to be static (hidden from the user) whereas the user can apply changes through `options`.

PR #634, opened 2019-01-08
Change cursors in Signaler to long to avoid overflowing

uedamm:
32-bit integer cursors in Signaler can overflow within a month or so under high load, and causes serious problems.

When wcursor in Signaler overflows, waitEvent(timeout == 0) cannot return true until recv() is called so that rcursor overflows as well. This condition is particularly harmful to IOThread, as it irreversibly renders IOThread unable to retrieve commands from its mailbox in inEvent(), because mailbox never calls any other methods in Signaler until it gets true from waitEvent() and IOThread always specifies timeout == 0 for mailbox.recv(). This in turn causes serious problems, such as a server being unable to accept new client connections anymore.

This pull request replaces 32-bit integer cursors with 64-bit long cursors. It does not fix the problem 100%, but fixes it within realistic durations.

### DIFF
--- a/src/main/java/zmq/Signaler.java
+++ b/src/main/java/zmq/Signaler.java
@@ -9,7 +9,7 @@ import java.nio.channels.Pipe;
 import java.nio.channels.SelectableChannel;
 import java.nio.channels.SelectionKey;
 import java.nio.channels.Selector;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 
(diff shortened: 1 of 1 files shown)


daveyarwood:
This looks reasonable to me, as a stopgap. :+1: 

I'm lacking context about this part of the codebase, but maybe a proper solution might look something like a mechanism for resetting the cursor to 0 occasionally, when it's safe to do that? Just spitballing, further discussion welcome.

Issue #641, opened 2019-02-06
wrong assert in curve mechanism


fbacchella:
I think this line https://github.com/zeromq/jeromq/blob/7d53a816e65e0a0243c69bd7a005a11c6ae1831f/src/main/java/zmq/io/mechanism/curve/CurveClientMechanism.java#L61
matches issue #605.

If someone set the private key but forget to set the public key, I'm afraid this line will throw a dirty exception. See [Options.setCurveKey](https://github.com/zeromq/jeromq/blob/master/src/main/java/zmq/Options.java#L600). Setting either public or private key is enough to activate a call to Curve mechanism.

fredoboulo:
That's a good finding!
I would keep the assert there, changing it to a pre-condition checking would result in another exception in a thread that is out of reach of the users. Maybe check these preconditions during bind and connect could allow better feedback to the calling code.

fbacchella:
Close as PR was merged.

PR #643, opened 2019-02-07
Adding a check of mechanisms settings.

fbacchella:
Try to resolve issue #641, by putting checks when binding or connecting for all security mechanisms.

### DIFF
--- a/src/main/java/zmq/io/mechanism/Mechanisms.java
+++ b/src/main/java/zmq/io/mechanism/Mechanisms.java
@@ -2,6 +2,8 @@ package zmq.io.mechanism;
 
 import java.nio.ByteBuffer;
 import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
 
 import zmq.Options;
 import zmq.ZMQ;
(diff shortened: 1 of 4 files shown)


src/main/java/zmq/io/mechanism/Mechanisms.java
```diff
@@ -36,6 +53,22 @@ public Mechanism create(SessionBase session, Address peerAddress, Options option
...
+        @Override
+        public void check(Options options)
+        {
+            if ((options.curvePublicKey == null ^ options.curveSecretKey == null)) {
+                throw new IllegalStateException("Curve mechanism definition incomplete");
+            }
+            if (options.curvePublicKey != null && options.curvePublicKey.length != Options.CURVE_KEYSIZE) {
+                throw new IllegalStateException("Curve mechanism definition incomplete");
+            }
+            if (options.curveSecretKey != null && options.curveSecretKey.length != Options.CURVE_KEYSIZE) {
```
fredoboulo: As a user, I would be welcoming all options config problems at once, the proposed algo means that if 2 keys are misconfigured, you will have to try 2 times to find out all the problems.

fbacchella:
openjdk build failed on , oraclejdk succeeded. It that really linked to my code ?

fredoboulo:
I got similar problem on my latest PR. Timeout on travis-ci after 10 mins. I assume it's not related to your change, but rather to infrastructure issues

daveyarwood:
In my experience, force-pushing can cause issues with Travis. The best thing to do when that happens is to retry the build in the Travis UI.

fredoboulo:
@daveyarwood where in Travis UI can we retry the build?

fbacchella:
perhaps raising this ZMQ.sleep(1):
https://github.com/zeromq/jeromq/blob/master/src/test/java/zmq/TestTermEndpoint.java#L236
would help.
Travis is very sensible to timing problem, as it can be very slow.

daveyarwood:
Sleeping in tests is an antipattern in general. Perhaps we should be checking the assertion in a loop and failing only if the assertion doesn't hold after a timeout (e.g. 5 seconds).

fredoboulo:
The test is all about disconnecting an endpoint and ensuring that it is not possible to send messages anymore. AND the disconnect process is asynchronous.
Best way we could do would be to monitor the socket and wait for the disconnection event... But it's not straightforward in zmq namespace.

trevorbernard:
This is a difficult problem since the disconnects are async and the monitor wasn't designed for flow/state control. It was a mechanism for logging and debugging.

daveyarwood:
I've written tests like this before -- I could take a stab at removing reliance on "sleep" in our tests sometime.

fbacchella:
Does it need to be in the critical path of this PR (or a 0.5 too) ?

fredoboulo:
No, it's not related to this PR. It just happened that it triggered the problem.

PR #646, opened 2019-02-08
Removing a really redundant stack trace, as the exception is returned.

fbacchella:
This stack trace is never lost, it's stored in a wrapping exception.
All the other one I found in this class needs special handling.


daveyarwood:
Good catch :+1:

PR #650, opened 2019-02-09
Clean up code in ZPoller using streams and lambdas

fbacchella:
Not really important, but cleaner code.


trevorbernard:
Can you rename the title to a problem your commit solves and I'd be happy to merge it in

PR #657, opened 2019-02-12
Removing some warning:

fbacchella:
- unused attributes.
- unused private methodes.
- annotation for the call of a necessary deprecated method.

### DIFF
--- a/src/main/java/org/zeromq/ZPoller.java
+++ b/src/main/java/org/zeromq/ZPoller.java
@@ -10,7 +10,6 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/zeromq/ZPoller.java
```diff
@@ -741,6 +741,7 @@ public void close() throws IOException
     /**
      * Destroys the poller without exception.
      */
+    @SuppressWarnings("deprecation")
```
daveyarwood: This deprecation warning is happening because we're calling `ZContext.closeSelector`, which is deprecated because the library user should never be managing selectors directly -- they are an implementation detail managed by the context.

I think instead of suppressing this warning, we should remove the call to `ctx.closeSelector` below and go back to `destroy()` having an empty body because there is nothing we need to do.

When a context is destroyed, it does exactly what we're doing here, so there's no need to do it here.

fbacchella: Ok, I didn't knew if it was really still used, so I keep it. I will remove it.

PR #658, opened 2019-02-12
Improved org.zeromq.ZMQ.Error enum

fbacchella:
Now all the declared error code in zmq.ZError are now in the Error enum.
It centralized the error message too, it make adding new messages easier.

### DIFF
--- a/src/main/java/org/zeromq/ZMQ.java
+++ b/src/main/java/org/zeromq/ZMQ.java
@@ -5,7 +5,9 @@ import java.nio.ByteBuffer;
 import java.nio.channels.SelectableChannel;
 import java.nio.channels.Selector;
 import java.nio.charset.Charset;
+import java.util.HashMap;
 import java.util.LinkedList;
+import java.util.Map;
 import java.util.Random;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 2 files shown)


src/main/java/org/zeromq/ZMQ.java
```diff
@@ -336,7 +338,7 @@ public static boolean proxy(Socket frontend, Socket backend, Socket capture, Soc
                              frontend.base,
                              backend.base,
                              capture == null ? null : capture.base,
-                             control == null ? null : control.base);
+                                             control == null ? null : control.base);
```
daveyarwood: This indentation looks strange -- I think the previous version is more correct, as we're lining up the method arguments vertically.

src/main/java/org/zeromq/ZMQ.java
```diff
@@ -447,18 +509,23 @@ public static void sleep(long amount, TimeUnit unit)
...
         {
-            for (Error e : Error.values()) {
-                if (e.getCode() == code) {
-                    return e;
-                }
+            if (map.containsKey(code)) {
+                return map.get(code);
+            }
+            else {
+                throw new IllegalArgumentException("Unknown " + Error.class.getName() + " enum code:" + code);
```
daveyarwood: could use a space here between the colon and the code:

```suggestion
                throw new IllegalArgumentException("Unknown " + Error.class.getName() + " enum code: " + code);
```

src/main/java/org/zeromq/ZMQ.java
fbacchella: My bad, this should not be in the comit.

src/main/java/org/zeromq/ZMQ.java
fbacchella: Indeed

PR #659, opened 2019-02-16
Improved concurrency of ZContext.

fbacchella:
It was using a CopyOnWriteArrayList, that is not a very efficient way of
managing concurrent access. It now uses a concurrent set, created with
ConcurrentHashMap.newKeySet(). So no manual locking is needed.

linger and sndhwm was protected with a lock, I think making them volatile
provides the same semantic (immediat change of value in all threads when set)
with less code.

It slightly changes the semantic of getSockets(). I'm not sure it's very
important.

In isClosed(), a totally useless synchronized(this) is removed.

### DIFF
--- a/src/main/java/org/zeromq/ZContext.java
+++ b/src/main/java/org/zeromq/ZContext.java
@@ -2,10 +2,10 @@ package org.zeromq;
 
 import java.io.Closeable;
 import java.nio.channels.Selector;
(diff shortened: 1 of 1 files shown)


fbacchella:
Change request done.

daveyarwood:
Great, thanks!

PR #680, opened 2019-02-27
Improved IPv6 and name resolution handling

fbacchella:
The actual way to handle IPv6 connectiont could be improved, a lot of code
had no effect. All the IPv6/IPv4 choice should be done in name resolution.

Error handling of failed name is improved too, a ZMQException is used.

Trying to cover many more test case in TCP name resolution.


src/main/java/zmq/Options.java
```diff
@@ -189,6 +189,9 @@ public Options()
         recvTimeout = -1;
         sendTimeout = -1;
         ipv6 = false;
+        if ("false".equals(System.getProperty("java.net.preferIPv4Stack", null)) || "true".equals(System.getProperty("java.net.preferIPv6Addresses", null))) {
+            ipv6 = true;
+        }
```
daveyarwood: Refactor idea: replace lines 191-194 with:

```java
ipv6 = "false".equals(System.getProperty("java.net.preferIPv4Stack", null)) ||
       "true".equals(System.getProperty("java.net.preferIPv6Addresses", null))
```

src/main/java/zmq/io/net/tcp/TcpAddress.java
```diff
@@ -114,24 +123,28 @@ public InetSocketAddress resolve(String name, boolean ipv6, boolean local)
...
-            for (InetAddress ia : InetAddress.getAllByName(addrStr)) {
-                if (ipv6 && !(ia instanceof Inet6Address)) {
-                    continue;
-                }
-                addrNet = ia;
-                break;
+            InetAddress[] addresses = InetAddress.getAllByName(addrStr);
+            if (ipv6) {
+                // prefer IPv6: return the first ipv6 or the first value if not found
+                addrNet = Arrays.stream(InetAddress.getAllByName(addrStr)).filter(i -> i instanceof Inet6Address).findFirst().orElseGet(() -> addresses[0]);
```
daveyarwood: Isn't `InetAddress.getAllByName(addrStr)` equivalent to `addresses` here?

src/main/java/zmq/io/net/tcp/TcpAddress.java
```diff
@@ -114,24 +123,28 @@ public InetSocketAddress resolve(String name, boolean ipv6, boolean local)
...
+            }
+            else {
+                addrNet = Arrays.stream(InetAddress.getAllByName(addrStr)).filter(i -> i instanceof Inet4Address).findFirst().orElseGet(() -> null);
```
daveyarwood: same question here

src/main/java/zmq/io/net/tcp/TcpConnecter.java
```diff
@@ -267,20 +267,9 @@ private boolean open() throws IOException
...
-            sa = resolved.address();
-            fd = SocketChannel.open();
-        }
-        assert (fd != null);
-
-        //  On some systems, IPv4 mapping in IPv6 sockets is disabled by default.
-        //  Switch it on in such cases.
+        // On some systems, IPv4 mapping in IPv6 sockets is disabled by default.
+        // Switch it on in such cases.
+        // Actually does nothing
```
daveyarwood: This comment (specifically the "does nothing" part) makes me very curious. It makes me think that maybe the code isn't doing what we're saying it's doing in the comments. Is there some context you could add to the comment?

src/main/java/zmq/io/net/tcp/TcpListener.java
```diff
@@ -174,14 +174,9 @@ public boolean setAddress(final String addr)
...
 
-            //  IPv6 address family not supported, try automatic downgrade to IPv4.
-            if (fd == null && address.family() == StandardProtocolFamily.INET6 && options.ipv6) {
-                // TODO V4 automatic downgrade to IPV4
-            }
-            assert (fd != null);
-
             //  On some systems, IPv4 mapping in IPv6 sockets is disabled by default.
             //  Switch it on in such cases.
+            // Actually does nothing
```
daveyarwood: same comment here re: the "does nothing" comment

src/main/java/zmq/io/net/tcp/TcpAddress.java
fbacchella: Indeed

src/main/java/zmq/io/net/tcp/TcpAddress.java
fbacchella: indeed again

src/main/java/zmq/io/net/tcp/TcpConnecter.java
fbacchella: In fact, the method TcpUtils.enableIpv4Mapping is empty. Actually does not have the same meaning in french, I will change that comment

src/main/java/zmq/io/net/tcp/TcpListener.java
fbacchella: again

src/main/java/zmq/io/net/tcp/TcpConnecter.java
fbacchella: But I wonder it that method make any sens at all. Isn't something that was copied directly from C code ? Java already wrap that protocol management. And I think it's too late when this code is reached.

daveyarwood: Nice -- thanks for the clarification!

I'm lacking the context to know if it makes sense for that method to exist or not... I think in general, it would be fine for us to diverge from the C code in places where Java handles parts of the implementation for us.

PR #688, opened 2019-03-10
Fixing javadoc.

fbacchella:
Allows remove -Xdoclint:none.


fbacchella:
Another travis failure. The jdk 11 branch is ok.

PR #691, opened 2019-03-12
Remove empty timer list

xinhochen:
During my work, I found failed connection retry will make the member 'timer' of PollerBase larger and larger, while the remaining list were all empty and would never be used. So remove them to reduce the memory usage.

### DIFF
--- a/src/main/java/zmq/poll/PollerBase.java
+++ b/src/main/java/zmq/poll/PollerBase.java
@@ -175,6 +175,15 @@ abstract class PollerBase implements Runnable
             timerInfo.sink.timerEvent(timerInfo.id);
         }
 
+        //  Remove empty list object
+        for (Entry<TimerInfo, Long> entry : timers.entries()) {
+            final Long key = entry.getValue();
+
+            if (timers.getValues(key).isEmpty()) {
+                timers.remove(key);
+            }
+        }
+
         if (changed) {
             return executeTimers();
         }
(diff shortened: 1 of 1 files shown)


daveyarwood:
This branch has a compilation error per Travis:

```
[ERROR] COMPILATION ERROR : 
[ERROR] /home/travis/build/zeromq/jeromq/src/main/java/zmq/poll/PollerBase.java:[182,23] getValues(K) has private access in zmq.util.MultiMap
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.5.1:compile (default-compile) on project jeromq: Compilation failure
[ERROR] /home/travis/build/zeromq/jeromq/src/main/java/zmq/poll/PollerBase.java:[182,23] getValues(K) has private access in zmq.util.MultiMap
```

fbacchella:
It could be written in a more java8 friendly way:

```
timers.entries().stream().filter(e -> ...).forEach(e -> timers.remove(e.getValue()));
```

trevorbernard:
```java
    private List<V> getValues(K key)
    {
        List<V> list = data.get(key);
        if (list == null) {
            list = new ArrayList<>();
            data.put(key, list);
        }
        return list;
    }
```
This will create a lot of garbage.

src/main/java/zmq/poll/PollerBase.java
```diff
@@ -175,6 +175,15 @@ protected long executeTimers()
             timerInfo.sink.timerEvent(timerInfo.id);
         }
 
+        //  Remove empty list object
+        for (Entry<TimerInfo, Long> entry : timers.entries()) {
+            final Long key = entry.getValue();
+
+            if (timers.getValues(key).isEmpty()) {
```
fredoboulo:
```
if (!timers.hasValues(key) {
   timers.remove(key);
}
```
could be possible.

xinhochen:
Thanks for reviewing, I've fixed it per note of @fredoboulo.

PR #692, opened 2019-03-12
Problem: ZBeacon does not provide the ability to specify server broadcast address.

### LINKED ISSUE #686: ZBeacon binds to wrong interface (or no option to specify broadcast interface)
When trying to broadcast beacons on a system with multiple network interfaces, broadcasts are not seen on all interfaces.
It appears that in a multi-network interface environment, ZBeacon will bind to an "automatically selected" interface (per java.nio.DatagramChannel docs). Subsequent broadcasts are only issued on that interface, potentially preventing it from broadcasting beacons on the correct network. Changing the "host" parameter in the ZBeacon constructor only changes the address to which broadcasts are broadcast to, but not the interface to which broadcasts are sent from.
ZBeacon should have a parameter that allows the user to specify the address (and thus interface) from which beacons should originate.

fredoboulo:
Solution: Create a builder to specify an optional broadcast address

Fixes #686

### DIFF
--- a/src/main/java/org/zeromq/ZBeacon.java
+++ b/src/main/java/org/zeromq/ZBeacon.java
@@ -11,20 +11,21 @@ import java.nio.channels.ClosedByInterruptException;
 import java.nio.channels.ClosedChannelException;
 import java.nio.channels.DatagramChannel;
 import java.util.Arrays;
+import java.util.Objects;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
 
 public class ZBeacon
 {
-    public static final long   DEFAULT_BROADCAST_INTERVAL = 1000L;
-    public static final String DEFAULT_BROADCAST_HOST     = "255.255.255.255";
+    public static final long    DEFAULT_BROADCAST_INTERVAL = 1000L;
+    public static final String  DEFAULT_BROADCAST_HOST     = "255.255.255.255";
+    private static final byte[] DEFAULT_BRODACAST_ADDRESS  = { 0, 0, 0, 0 };
 
-    private final int             port;
-    private InetAddress           broadcastInetAddress;
-    private final BroadcastClient broadcastClient;
-    private final BroadcastServer broadcastServer;
-    private final byte[]          beacon;
-    private byte[]                prefix            = {};
-    private long                  broadcastInterval = DEFAULT_BROADCAST_INTERVAL;
-    private Listener              listener          = null;
+    private final BroadcastClient           broadcastClient;
+    private final BroadcastServer           broadcastServer;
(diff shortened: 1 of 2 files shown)


fredoboulo:
I'm puzzled here: compilation errors on ByteBuffer method, full tests suite is passing locally fo rme.
Revert?

fbacchella:
Is that this one: 
https://github.com/openhab/openhab1-addons/issues/5663#issuecomment-425674686 ?
I get bitten by that: https://github.com/rrd4j/rrd4j/issues/119

fredoboulo:
Yep: https://github.com/apache/felix/pull/114. It stings!
My JVM is now 10.0.2, so problem was not detected on my machine.
Great, now I have to reinstall Java8 ...

fbacchella:
You can have a look at https://github.com/fbacchella/HpSmu/blob/master/pom.xml and https://github.com/fbacchella/HpSmu/blob/master/.circleci/config.yml the way I found to solve that. It allows to build using up to date JVM (and so compiler), but the required java runtime:

in pom:
```
    <properties>
        <maven.compiler.source>1.8</maven.compiler.source>
        <maven.compiler.target>1.8</maven.compiler.target>
        <jdk.compile.home>${java.home}</jdk.compile.home>
    </properties>
...
    <build>
        <plugins>
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-compiler-plugin</artifactId>
                <configuration>
                    <compilerArgs>
                        <arg>-Xbootclasspath:${jdk.compile.home}/lib/rt.jar</arg>
                    </compilerArgs>
                </configuration>
            </plugin>
```

And it's launched with:
```
JAVA_HOME=$java11 mvn compile -Djdk.compile.home=$java8/jre
```

src/main/java/org/zeromq/ZBeacon.java
```diff
@@ -11,20 +11,21 @@
...
-    private final int             port;
-    private InetAddress           broadcastInetAddress;
-    private final BroadcastClient broadcastClient;
-    private final BroadcastServer broadcastServer;
-    private final byte[]          beacon;
-    private byte[]                prefix            = {};
-    private long                  broadcastInterval = DEFAULT_BROADCAST_INTERVAL;
-    private Listener              listener          = null;
+    public static final long    DEFAULT_BROADCAST_INTERVAL = 1000L;
+    public static final String  DEFAULT_BROADCAST_HOST     = "255.255.255.255";
```
fbacchella: How that does handle IPv6 ?

fredoboulo: Default is IPv4 format, users wanting to use IPv6 can do it by configuring the client host; using one of the constructors ZBeacon(String host, ...), or the new builder.

PR #710, opened 2019-04-12
Problem: Usages of some java API forbid usage in Android

### LINKED ISSUE #647: Jeromq latest snapshot crashes when used on Android, where as 0.4.3 works fine
Jeromq latest snapshot crashes when used on Android, where as 0.4.3 works fine. I have attached an android app project. Importing this and running it on android 8 emulator can reproduce the issue. I tried debugging. It is happening in the TcpListener.java in setAddress() method while binding and causing a crash in libc. If any one wants to take a look, I can help with the setup as well.
https://drive.google.com/file/d/1bFdu6VFeAwq1gRZK3OnaWZhAi-_YksQN/view?usp=sharing

fredoboulo:
Solution: Remove usages of streams and default to traditional loops.

Fixes #647

### DIFF
--- a/src/main/java/zmq/io/net/tcp/TcpAddress.java
+++ b/src/main/java/zmq/io/net/tcp/TcpAddress.java
@@ -6,7 +6,6 @@ import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
 import java.net.UnknownHostException;
-import java.util.Arrays;
 
 import org.zeromq.ZMQException;
 
@@ -133,10 +132,23 @@ public class TcpAddress implements Address.IZAddress
             InetAddress[] addresses = InetAddress.getAllByName(addrStr);
             if (ipv6) {
                 // prefer IPv6: return the first ipv6 or the first value if not found
-                addrNet = Arrays.stream(addresses).filter(i -> i instanceof Inet6Address).findFirst().orElseGet(() -> addresses[0]);
+                for (InetAddress addr : addresses) {
+                    if (addr instanceof Inet6Address) {
+                        addrNet = addr;
+                        break;
+                    }
+                }
+                if (addrNet == null) {
+                    addrNet = addresses[0];
+                }
             }
             else {
-                addrNet = Arrays.stream(addresses).filter(i -> i instanceof Inet4Address).findFirst().orElseGet(() -> null);
+                for (InetAddress addr : addresses) {
+                    if (addr instanceof Inet4Address) {
+                        addrNet = addr;
+                        break;
+                    }
+                }
             }
         }
         catch (UnknownHostException e) {


daveyarwood:
@fredoboulo I think we should add the problematic packages/methods to the checkstyle config like you did in #702, to keep us from making the same mistake in the future.

fredoboulo:
The problematic usage is calling stream methods. These methods are spread in various classes. And I couldn't find ootb checker that forbids some method calls (I did not investigate much though).
I can see some alternatives: 
Create a pattern that forbids ".stream("
Create our own checker
Use other tools: findbugs might help.
Have a travis android configuration.
The main point for me is that we can only react to feedbacks of users stating it does not work. That happens usually after we release.

fbacchella:
This lists the differences:
https://developer.android.com/studio/write/java8-support.
Does this tools works: http://tools.android.com/tips/lint ?
SonarCloud can take Android Lint reports

fredoboulo:
Yes but this list does not include Arrays.stream or Collection.stream. that's kind of obvious as an afterthought but we fell into that trap

daveyarwood:
Would it make sense to start adding individual methods like Arrays.stream and Collection.stream to the blacklist?

trevorbernard:
Just add the appropriate android envs to the CI/CD env. If it breaks then we don't merge. Fully compatible with c4. But we need to figure out which android env we want to support. E.g. 4.4 aka kitkat is incredibly old and likely not something we want to support.

fredoboulo:
I prefer that, it would allow to make a statement like Jeromq supports android up to API level ... And we are removing the risk of forgetting that one method/class/... is not implemented for our minimum level of compatibility. Next step would be indeed to enhance checkstyle rules for the contributors to be informed of breaking changes even before submitting a PR.

daveyarwood:
I like it too. I'm actually about to submit a PR to (attempt to) add an Android build to our Travis setup.

fredoboulo:
I would propose level 19 as a minimum. That gives 96ish % of compatibility,  and that can only increase in the future.

trevorbernard:
I've been meaning to show the circleci build some love but I've just had higher priority things in my life

daveyarwood:
@trevorbernard Slightly off-topic, but do you have any thoughts about Travis vs. CircleCI? Do you think we might want to move over to just Circle at some point, or is the idea that we have both?

trevorbernard:
I think circleci is the right direction because you can add pipelines and the future of travisci is uncertain after they got picked up by a private equity firm. But I haven't thought about it in any detail

fredoboulo:
A little bit of digging:

https://developer.android.com/reference/java/util/Map.html#computeIfAbsent(K,%20java.util.function.Function%3C?%20super%20K,%20?%20extends%20V%3E)

"Added in API level 24"

Providing an exhaustive list of methods that cannot be used will be a long-running task. An interesting one if contributors can have guidance before submitting the PR, but that exposes the risk of not being exhaustive.

PR #711, opened 2019-04-13
Travis config tweaks

daveyarwood:
Continuing the conversation from #710, it would be nice if we had an Android build on Travis. I've started poking around a bit at the Travis documentation, so here is an attempt.

I will probably tinker with this PR a bit, trying things out and seeing how the builds go. Please hold off on merging for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,20 @@
+matrix:
+  include:
+    - language: java
+      jdk:
+        - openjdk8
+        - oraclejdk8
+
+    - language: android
+      jdk:
+        - openjdk8
+        - oraclejdk8
+      android:
+        components:
+          - tools
+          - platform-tools
+
 install: mvn install --quiet -DskipTests=true -Dgpg.skip=true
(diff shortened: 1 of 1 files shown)


daveyarwood:
References:

https://docs.travis-ci.com/user/languages/android/
https://docs.travis-ci.com/user/build-matrix/
https://stackoverflow.com/a/44054333/2338327

daveyarwood:
I'm attempting to just use the latest revision of the Android tools for now and see if we can get that working. It might make sense to add older versions of the Android SDK and build tools to the config, but I'm pretty far out of the loop when it comes to Android, so I'll leave that as a TODO item.

daveyarwood:
Both the Java and the Android builds passed! :tada: 

One wrinkle is that it only ran Java w/ OpenJDK 8 and Android w/ OpenJDK 8, and didn't run either of them with Oracle Java 8. I think I know why that is. Will submit a new PR to continue tinkering.

PR #714, opened 2019-04-13
Speed up Travis build by caching Maven dependencies

daveyarwood:
This should make our Travis builds faster -- should be interesting to see by how much.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,10 @@ matrix:
           - tools
           - platform-tools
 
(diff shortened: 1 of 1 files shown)


daveyarwood:
Reference: https://docs.travis-ci.com/user/caching/

daveyarwood:
NB: This first build will be as slow as always, since we don't have a cache yet. But then after that, builds should be faster.

PR #731, opened 2019-06-16
use simpler java collection instead of array in internal of Poller class

esahekmat:
it's more readable and simpler to work with List instead of an array. it removes all the resizing and size management stuff

### DIFF
--- a/src/main/java/org/zeromq/ZMQ.java
+++ b/src/main/java/org/zeromq/ZMQ.java
@@ -7,6 +7,7 @@ import java.nio.channels.Selector;
 import java.nio.charset.Charset;
 import java.util.HashMap;
 import java.util.LinkedList;
+import java.util.List;
 import java.util.Map;
 import java.util.Random;
 import java.util.concurrent.TimeUnit;
@@ -3614,14 +3615,9 @@ public class ZMQ
         private final Selector selector;
         private final Context  context;
 
-        private PollItem[] items;
-        private int        next;
-        private int        used;
+        private List<PollItem> items;
 
         private long timeout;
-
-        // When socket is removed from polling, store free slots here
-        private final LinkedList<Integer> freeSlots;
 
         /**
          * Class constructor.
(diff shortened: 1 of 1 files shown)


src/main/java/org/zeromq/ZMQ.java
```diff
@@ -3637,11 +3633,8 @@ protected Poller(Context context, int size)
             selector = context.selector();
             assert (selector != null);
 
-            items = new PollItem[size];
+            items = new LinkedList<>();
```
fbacchella: As size is now useless, the constructor using a size should be deprecated. Or size could be used when creating the list, it avoid some useless resizing.

colini:
I'm wondering why LinkedList instead of ArrayList?

fbacchella:
If you add and remove many elements, it will do less reallocation of full array. But as the real key for search is the socket, not the index (a legacy from C code probably), I wonder if using a map could be better ? With a deprecation added for all index-based access.

src/main/java/org/zeromq/ZMQ.java
esahekmat: resizing is cheap on LinkedList, but you're right, I should deprecate this constructor and create a new one without size

esahekmat:
exactly, I chose LinkedList because its resizing is cheap, map sounds good, it's faster for search which is the used frequently

esahekmat:
Actually, I found out that methods like poller.pollout(index) and poller.pollin(index) is more frequently used over and over usually in a while loop, and there are no similar methods getting Socket instead of index, I can add these methods and use a map instead of LinkedList the only problem is Poller accept SelectableChannel also, and we should use Object as the key of the map.

trevorbernard:
Subjectively, I like array backed data structures more because they're more cache friendly because you aren't jumping from pointer to pointer especially in a hot loop. But in any case, it's a good commit so it will be merged.

esahekmat:
Thank you @trevorbernard, from your comment I understand that it's better to use ArrayList to avoid jumping from pointer to pointer. so let me open another pull request in order to change LinkedList to ArrayList

PR #733, opened 2019-06-21
fix #729

### LINKED ISSUE #729: ZPoller.poll returning -1 with v0.5.1
Jeromq v0.5.1
MacOSX 10.14.4

```
var inPoller = new ZPoller(context);
inPoller.register(new ZPoller.ZPollItem(inprocSocket, null, ZPoller.POLLIN));
int inPollEventsCount = inPoller.poll(-1);
```

Not sure what happened between the upgrade, but I get 1 on poll on v0.5.0, but -1 on v0.5.1, so something went wrong somewhere. Something undocumented?

esahekmat:
This test simulates issue #729
This test passes on tag: v0.5.0 but it fails now, As I understand it's related to the newly introduced class CompositePollItem, and pass a list of its instance instead of the list of ItemHolders to poll method cause this problem, @fredoboulo would you help me to solve this problem?

### DIFF
--- a/src/test/java/org/zeromq/TestZPoller.java
+++ b/src/test/java/org/zeromq/TestZPoller.java
@@ -518,7 +518,8 @@ public class TestZPoller
             assertTrue(pollin);
             String hello = sub.recvStr();
             assertThat("recieved message are not identical to what has been sent", hello, is("hello"));
-        } finally {
+        }
+        finally {
             zPoller.close();
             ctx.close();
         }
--- a/src/test/java/org/zeromq/TestZPoller.java
+++ b/src/test/java/org/zeromq/TestZPoller.java
@@ -491,30 +491,36 @@ public class TestZPoller
         boolean subscribe = sub.subscribe("");
         assertTrue("SUB Socket could not subscribe", subscribe);
         ZPoller zPoller = new ZPoller(ctx);
-        zPoller.register(new ZPoller.ZPollItem(sub, null, ZPoller.POLLIN));
 
-        Thread server = new Thread(new Runnable()
-        {
-            @Override
-            public void run()
+        try {
+            zPoller.register(new ZPoller.ZPollItem(sub, null, ZPoller.POLLIN));
+
+            Thread server = new Thread(new Runnable()
             {
-                while (true) {
-                    pub.send("hello");
-                    try {
-                        Thread.sleep(100);
-                    }
-                    catch (InterruptedException ignored) {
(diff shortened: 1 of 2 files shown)


fredoboulo:
Mmmm, there are no handlers at all, can you try to change ZPoller, with something

```
diff --git a/src/main/java/org/zeromq/ZPoller.java b/src/main/java/org/zeromq/ZPoller.java
index aa330e5..e234715 100644
--- a/src/main/java/org/zeromq/ZPoller.java
+++ b/src/main/java/org/zeromq/ZPoller.java
@@ -320,16 +320,13 @@ public class ZPoller implements Closeable
             boolean first = true;
             for (ItemHolder holder : holders) {
                 if (holder.item().hasEvent(events)) {
+                    if (first) {
+                        first = false;
+                        has = true;
+                    }
                     EventsHandler handler = holder.handler() == null ? globalHandler : holder.handler();
                     if (handler != null) {
-                        boolean evts = handler.events(socket, events);
-                        if (first) {
-                            first = false;
-                            has = evts;
-                        }
-                        else {
-                            has &= evts;
-                        }
+                        has &= handler.events(socket, events);
                     }
                 }
             }
```

esahekmat:
your fix make it pass, thank you

src/test/java/org/zeromq/TestZPoller.java
```diff
@@ -477,4 +478,43 @@ private final Pipe pipe(ZPoller poller, BiFunction<SelectableChannel, Integer, B
 
         return pipe;
     }
+
+    @Test
+    public void testIssue729() throws InterruptedException, IOException
+    {
+        int port = Utils.findOpenPort();
+        ZContext ctx = new ZContext();
```
fredoboulo: context needs to be closed

src/test/java/org/zeromq/TestZPoller.java
```diff
@@ -477,4 +478,43 @@ private final Pipe pipe(ZPoller poller, BiFunction<SelectableChannel, Integer, B
...
+        Socket sub = ctx.createSocket(SocketType.SUB);
+        Socket pub = ctx.createSocket(SocketType.PUB);
```
fredoboulo: sockets need to be closed

src/test/java/org/zeromq/TestZPoller.java
esahekmat: I closed poller and context

PR #735, opened 2019-07-14
Fix #638: remove data field from zmq.Msg

### LINKED ISSUE #638: jeromq is adding 7 bytes when frame size is large(~MBs) on Android
@trevorbernard @markif 
When the message has multiple frames, and if any of the frame size is larger(I noticed it when I am sending around 4MB). It is adding exactly 7 bytes. Although Msg class keeps track of size, but doesn't limit the buffer size before passing it on to ZFrame. I am using a workaround by calling recv and building ZMsg myself.
<del>
```
 private zmq.Msg recv(ZMQ.Socket socket, int flags)
    {
        zmq.Msg msg = socket.base().recv(flags);

        if (msg != null) {
            return msg;
        }
        return null;
    }

    private ZMsg recvMsg(ZMQ.Socket socket, int flag)
    {
        if (socket == null) {
            throw new IllegalArgumentException("socket is null");
        }

        ZMsg msg = new ZMsg();

        while (true) {
            zmq.Msg internalMsgFrame = recv(socket, flag);
            ByteBuffer buf = internalMsgFrame.buf();
            if (buf == null) {
                return null;
            }
            if (internalMsgFrame == null) {
                // If receive failed or was interrupted
                msg.destroy();
                msg = null;
                break;
            }
            // check size and resize;
            buf.limit(internalMsgFrame.size());
            byte[] data = new byte[buf.remaining()];
            buf.duplicate().get(data);
            msg.addLast(data);
            if (!(socket.base().getSocketOpt(zmq.ZMQ.ZMQ_RCVMORE) == 1)) {
                break;
            }
        }
        return msg;
    }
```
</del>
**NOTE:** The above workaround is wrong as android adds 4bytes in beginning and 3 bytes at end.

knuesel:
As discussed in https://github.com/zeromq/jeromq/issues/638#issuecomment-508792276 the bug of wrong buffers returned by `Msg#recv` can be fixed by removing the (mostly redundant) `data` field from the `Msg` class.

The bug was found with buffers allocated through `ByteBuffer.allocateDirect()` on Android, but the same problems occur with regular buffers obtained from `ByteBuffer#slice()` (any buffer that doesn't fully use its backing array can cause problem). The proposed change includes 5 tests that fail in the current version but are fixed here (the `slice()` method is used to emulate the Android buffer allocation).

This also fixes the bug where `Msg#get()` and `Msg#getBytes()` give inconsistent results (see https://github.com/zeromq/jeromq/issues/638#issuecomment-508792276 ).

Note that the `Msg(final Msg) m` constructor copies the ByteBuffer of the `m` instance without copying the underlying data (so both messages will share the same data). It seems strange to me, but maybe it is the desired behavior? (I didn't change it.)

### DIFF
--- a/src/main/java/zmq/Msg.java
+++ b/src/main/java/zmq/Msg.java
@@ -228,7 +228,8 @@ public class Msg
             if (buf.arrayOffset() == 0 && array.length == size) {
                 // If the backing array is exactly what we need, return it without copy.
                 return array;
-            } else {
+            }
+            else {
                 // Else use it to make an efficient copy.
                 return Arrays.copyOfRange(array, offset, offset + size);
             }
@@ -355,7 +356,8 @@ public class Msg
 
         if (buf.hasArray()) {
             System.arraycopy(buf.array(), buf.arrayOffset() + index, dst, off, count);
-        } else {
+        }
+        else {
             ByteBuffer dup = buf.duplicate();
             dup.position(index);
             dup.get(dst, off, count);
--- a/src/main/java/zmq/Msg.java
+++ b/src/main/java/zmq/Msg.java
@@ -4,6 +4,7 @@ import java.io.ByteArrayOutputStream;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.nio.channels.SocketChannel;
+import java.util.Arrays;
 
 import zmq.io.Metadata;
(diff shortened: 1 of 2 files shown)


knuesel:
@yagnasrinath glad if it helps!

PR #767, opened 2020-02-26
Fix a bug when ticket timeout less than zero

imkcy9:
Fix a bug when ticket timeout less than zero


imkcy9:
> Could you add some unit tests?
> I'm curious how you encountered that problem.

Hi. This is an unit test. If a ticker is time out, it is supposed to return 0 the same as a timer does.

PR #775, opened 2020-04-18
problem: ZMQ_HEARTBEAT is not useful without sending an hello message

somdoron:
When using ZMQ_HEARTBEAT one still needs to implement application-level heartbeat in order to know when to send a hello message.
For example, with the majordomo protocol, the worker needs to send a READY message when connecting to a broker. If the connection to the broker drops, and the heartbeat recognizes it the worker won't know about it and won't send the READY msg.
To solve that, the majordomo worker still has to implement heartbeat. With this new option, whenever the connection drops and reconnects the hello message will be sent, greatly simplify the majordomo protocol, as now READY and HEARTBEAT can be handled by zeromq.

Port of https://github.com/zeromq/libzmq/pull/3870

### DIFF
--- a/src/main/java/org/zeromq/ZMQ.java
+++ b/src/main/java/org/zeromq/ZMQ.java
@@ -2989,6 +2989,20 @@ public class ZMQ
         }
 
         /**
+         * When set, the socket will automatically send a hello message when a new connection is made or accepted.
+         * You may set this on DEALER or ROUTER sockets.
+         * The combination with ZMQ_HEARTBEAT_IVL is powerful and simplify protocols,
+         * as now heartbeat and sending the hello message can be left out of protocols and be handled by zeromq.
+         *
+         * @param helloMsg
+         * @return true if the option was set, otherwise false
+         */
+        public boolean setHelloMsg(byte[] helloMsg)
+        {
+            return setSocketOpt(zmq.ZMQ.ZMQ_HELLO_MSG, helloMsg);
+        }
+
+        /**
          * Bind to network interface. Start listening for new connections.
          *
          * @param addr the endpoint to bind to.
--- a/src/main/java/zmq/Ctx.java
+++ b/src/main/java/zmq/Ctx.java
@@ -729,6 +729,13 @@ public class Ctx
             assert (written);
             pendingConnection.bindPipe.flush();
         }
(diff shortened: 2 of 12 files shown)


RalphSteinhagen:
@somdoron this is not an issue but more a follow-up question. Saw that you are not on Gitter but that you contributed significantly on this (the above) topic in JeroMQ. Would you be able to confirm/deny my 'connection heartbeating' hypothesis/question I posted at Gitter [here](https://gitter.im/zeromq/general?at=604b2fd422a5ce4a914e2625)?

Any help or even brief 'yes'/'no' would be much appreciated.

> Has someone experiences with JeroMQ and '[Connection Heartbeating](https://rfc.zeromq.org/spec/37/#connection-heartbeating)'? 
> 
> We already use application-level heartbeating but I was curious whether the following could help/simplify this. The [README.md](https://github.com/zeromq/jeromq#user-content-features) states that it follows the [ZMTP/3.0](http://rfc.zeromq.org/spec:23) but heartbeating seems to be part of [ZMTP/3.1](https://rfc.zeromq.org/spec/37/#connection-heartbeating). How much of this is already working in JeroMQ? 
> 
> The spec defines only one 'ping-ttl' timeout which I guess is the one mapped to `Socket::setHeartbeatTtl(int)` (A). What is the difference w.r.t. 'Socket::setHeartbeatTimeout(int)' (B)? Is this an asymmetric time-out definition ala
> * 'if you do not respond to my ping after A ms, I consider you offline' and 
> * 'you may consider me offline, if I haven't sent any messages/pings within B ms'
> and `Socket::setHeartbeatTtl(int)` setting the frequency of the transparent pings that my client sends under the hood?
> 
> For example if I choose the following socket parameters:
> 
> ```
> socket.setHeartbeatTtl(1000);
> socket.setHeartbeatTimeout(3000);
> socket.setHeartbeatIvl(200);
> ```
> 
> Does this imply, that my client sends pings every 200 ms, if the other client/server does not reply within 1000 ms my client socket closes, and if I haven't sent any (new) pings for 3000 ms the other client/server may (or is forced to) close the socket on it's side? Does this work for all tcp-based socket combinations?
> 
> Any clarification/confirmation would be much appreciated.

PR #777, opened 2020-04-20
Resolve #776.

### LINKED ISSUE #776: ZContext.destroySocket might forget socket
This method says:
```
    public void destroySocket(Socket s)
    {
        if (s == null) {
            return;
        }
        s.setLinger(linger);
        s.close();
        this.sockets.remove(s);
    }
```

It the thread is interrupted, s.close() will throws a zmq.ZError$IOException, but before the sockets.remove(s);

So the socket is closed anyway because of the `isClosed.compareAndSet(false, true)` in ZMQ.Socket.close() code, but it's still in  ZContext.sockets Set. So I think the code could be:
```
try {
    s.close();
}  finally {
    this.sockets.remove(s);
}
```

fbacchella:
If a socket failed during close, it's not removed from ZContext, even
if it should.

### DIFF
--- a/src/main/java/org/zeromq/ZContext.java
+++ b/src/main/java/org/zeromq/ZContext.java
@@ -162,8 +162,12 @@ public class ZContext implements Closeable
             return;
         }
         s.setLinger(linger);
-        s.close();
-        this.sockets.remove(s);
+        try {
+            s.close();
+        }
+        finally {
+            sockets.remove(s);
+        }
     }
 
     /**


daveyarwood:
Looks like there is a checkstyle violation:

> [ERROR] src/main/java/org/zeromq/ZContext.java:[167,9] (blocks) RightCurly: '}' at column 9 should be alone on a line.

fbacchella:
The build succeed on my local machin. I'm not sure current failures are coming from this code.

fredoboulo:
It seems to be unrelated, CircleCI was always failing for me. And there is one failing build out of 7 in Travis, where a totally unrelated test was stuck

daveyarwood:
Agreed - this seems to just be a flaky test.

PR #778, opened 2020-04-21
ZMQ.Socket now remember the ZContext that created it

fbacchella:
The ZContext used to create a ZMQ.Socket is now stored within the socket. So with this patch, socket.close() is now enough to remove the socket from the internal state of ZContext.

It makes the use of try-with-resources much easier.

It also remove the unused Ctx storage from  ZMQ.Socket.


daveyarwood:
The code change looks good, but all of the Travis builds are failing, so I think there must be some subtle bug?

daveyarwood:
Only 3 of the builds are failing now, timing out on various test cases in ZAuthTest. I think this is unrelated to the PR, so I'm going to go ahead and merge it.

PR #781, opened 2020-04-22
Javadoc on recent javahome fails.

fbacchella:


daveyarwood:
Hmm, there are a ton of Javadoc-related errors in the Travis build on this branch.

fbacchella:
It tried it I home. It's failing with any java version previous that 13. So I think that I need to fix that as both 1.8 and 11 are LTS.

fbacchella:
OK, I have one incompatible change, I need to choose if it fails on latest java (13+) or earlier. Since jeromq aim java 1.8 and Android, I choose the 1.8 compatible version.
I have a setup were I can build java-1.8 compatible jar with the latest jvm (14). But it needs a little setup in the pom and break out-of-the box compilation. An example can be found at https://github.com/rrd4j/rrd4j/blob/master/README.md#building-optional. The CI setup is complicated too.

PR #783, opened 2020-05-08
problem: jeromq is not thread-safe

somdoron:
Solution: port CLIENT and SERVER sockets from libzmq, which are thread-safe sockets.

The thread-safe sockets family includes other sockets, which are not part of this commit.

Also, CLIENT and SERVER cannot be used with a poller at the moment. For poller support, zmq_poller from libzmq has to be ported as well.

### DIFF
--- a/src/main/java/zmq/MailboxSafe.java
+++ b/src/main/java/zmq/MailboxSafe.java
@@ -1,0 +1,134 @@
+package zmq;
+
+import zmq.pipe.YPipe;
+import zmq.util.Errno;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.concurrent.locks.Condition;
+
+public class MailboxSafe implements IMailbox
+{
+    //  The pipe to store actual commands.
+    private final YPipe<Command> cpipe;
+
+    //  Synchronize access to the mailbox from receivers and senders
+    private final ReentrantLock sync;
+
+    //  Condition variable to pass signals from writer thread to reader thread.
+    private final Condition condition;
+
+    private final ArrayList<Signaler> signalers;
+
+    // mailbox name, for better debugging
+    private final String name;
+
+    private final Errno errno;
+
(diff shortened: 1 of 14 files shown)


trevorbernard:
Awesome stuff.. I'll check it out this evening.

src/main/java/zmq/MailboxSafe.java
```diff
@@ -0,0 +1,131 @@
...
+    //  The pipe to store actual commands.
+    private final YPipe<Command> cpipe;
+
+    //  Synchronize access to the mailbox from receivers and senders
+    private final ReentrantLock sync;
+
+    //  Condition variable to pass signals from writer thread to reader thread.
+    private final Condition condition;
+
+    private final Vector<Signaler> signalers;
```
trevorbernard, line 23: Vectors are synchronized. If thread safety isn't needed consider using an ArrayList

src/main/java/zmq/MailboxSafe.java
```diff
@@ -0,0 +1,131 @@
...
+    public void send(Command cmd)
+    {
+        sync.lock();
+        try {
+            cpipe.write(cmd, false);
+            boolean ok = cpipe.flush();
+
+            if (!ok) {
+                condition.signalAll();
+                signalers.forEach(Signaler::send);
```
trevorbernard, line 71: Java iterators generate garbage unfortunately. Not sure if it matters here

somdoron: it matters, this would be called alot, is this also generate garbage:

```
for (Signaler signaler : signalers) {
    signaler.send();
}
```

or should I use:

```
for (int i = 0; i < signalers.size(); i++) {
    signalers.get(i).send();
}
```

src/main/java/zmq/MailboxSafe.java
somdoron: fixed

src/main/java/zmq/MailboxSafe.java
trevorbernard: the latter

PR #801, opened 2020-05-17
Another proposition for handling termination of ZContext.

fbacchella:
I sill think that #destroySocket should be deprecated. I don't see it's use case.

The default linger is applied on context shutdown. I tried to improved the
javadoc of #setLinger.


fbacchella:
Wrong approach,  trying something else.

Issue #800, opened 2020-05-15
Failure of org.zeromq.TestZPoller.testPollerPollout


fbacchella:
While investigating the many failure of this test, I found something that can be seen as a bug.
This test works by having a thread sending a single message and then quitting, see https://github.com/zeromq/jeromq/blob/master/src/test/java/org/zeromq/TestZPoller.java#L108.
This event is often lost.
But if I ensure that the sending socket is closed after the receiving socket, it succeed. 
I added a single CountDownLatch.await() right before the break . This latch is release on the other side, right after the poll (https://github.com/zeromq/jeromq/blob/master/src/test/java/org/zeromq/TestZPoller.java#L165). That's something very strange and it look like more a bug that a test instability. So this latch is more a workaround than a real solution and might explains many of the tests failures.

fredoboulo:
... It's a PUSH-PULL there, and the immediate flag is not set. Can you try setting it on the PUSH socket before binding/connecting?

fbacchella:
It didn't help.

fredoboulo:
I made a git bisect for that particular test, seems that 

> #4e0d84093df8bd148510b6dd3851bb535894900a
>  ZMQ.Socket now remember the ZContext that created it and remove from it when closed.

is the first commit where the test hangs.

When I changed org.zeromq.ZMQ.Socket#close to call directly internalClose, it works.
From what I see, calling destroySocket from the ZContext sets the linger beforehand to 0, meaning deactivating the linger, and not sending the pending messages in queue. This was confirmed by setting `context.setLinger(-1);` in org.zeromq.TestZPoller#testPollerPollout, which makes the test pass without any other modification.

What worries me is that it breaks the previously released behaviour: Only ZContext sets the linger to 0 before destroying the context, ZMQ.Socket does not and uses the linger of the socket, which is -1 by default.

fbacchella:
The problem seems to be that ZContext#destroySocket is not a smart close. It indeed does what it says: destroying the socket in emergency, with a short linger. So Socket#close should not call it, but a weaker ZContext#socketClose that just remove the reference.
What should ZContext#close should call ? Should it stay with destroySocket ?

fbacchella:
I merged a correction in #45ba936f6ea83b3be0b8cdf1d684fde347f85bf0. Sorry for the direct merge, it was not my intention.

trevorbernard:
I believe this was resolved with #803. If not feel free to reopen

PR #802, opened 2020-05-18
Third try to improve TestZPoller.

fbacchella:
Issue #800  was indeed a problem in the linger of the test. But I also raising problems
with linking of linger in ZMQ.Socket and ZContext. When a socket is obtained
from a ZContext, the linger value of this socket is meaningless. So let
ZContext handle it for all it's owned socket, and explained it in the documentation.

That don't change the old inner working, but make it more explicit in code
and documentation.

Another solution would be to use the context linger only as an hint and let
each socket manage it's own linger, so removing the Socket.setLinger in
ZContext.destroySocket, just using it at socket creation.

Any way, in the test, the linger needs to be set to a non null value.

### DIFF
--- a/src/main/java/org/zeromq/ZContext.java
+++ b/src/main/java/org/zeromq/ZContext.java
@@ -162,26 +162,6 @@ public class ZContext implements Closeable
         if (s == null) {
             return;
         }
-        s.setLinger(linger);
-        try {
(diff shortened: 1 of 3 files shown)


trevorbernard:
Deprecated by #803

PR #805, opened 2020-05-18
Fix #804 - Change setBeacon to be synchronized

### LINKED ISSUE #804: Problem: setBeacon in ZBeacon isn't thread safe
getBeacon is thread safe where as setBytes isn't.

trevorbernard:


trevorbernard:
The methods shouldn't be synchronized at all

PR #812, opened 2020-05-19
NetProtocol was not made Enum enough

fbacchella:
- Protocol name was used in place were NetProtocol would be better.
- Using .equals() were == works and is more readable.
- Remove some tests where a simple provider method in enum can do the same with more straigthfull code
- Adding a Listener abstract class, to hide the actual implementation of ipc protocol using tcp.
- NetProtocol.getProcol throws an IllegalArgumentException, more robust than a null that can sneak around. So SocketBase.checkProtocol needs to handle that.
- zmq.io.net.Address was missing a constructor using directly NetProtocol, deprecate the old constructor taking a String.
- Added explicit flag for multicasp transport kind. Is it redudant with subscribe2all ?
- Slightly better NetProtocol.isCompatible, using a Set instead of an List.

### DIFF
--- a/src/main/java/zmq/io/net/Listener.java
+++ b/src/main/java/zmq/io/net/Listener.java
@@ -1,0 +1,23 @@
+package zmq.io.net;
+
+import zmq.Options;
+import zmq.Own;
+import zmq.SocketBase;
+import zmq.io.IOThread;
+import zmq.poll.IPollEvents;
+
+public abstract class Listener extends Own implements IPollEvents
+{
+    //  Socket the listener belongs to.
+    protected final SocketBase socket;
+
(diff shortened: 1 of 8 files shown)


src/main/java/zmq/io/net/Listener.java
```diff
@@ -0,0 +1,23 @@
...
+
+import zmq.Options;
+import zmq.Own;
+import zmq.SocketBase;
+import zmq.io.IOThread;
+import zmq.poll.IPollEvents;
+
+public abstract class Listener extends Own implements IPollEvents
+{
+    //  Socket the listerner belongs to.
```
trevorbernard, line 11: Typo

fbacchella: Fixed

PR #833, opened 2020-06-05
Fix flaky ReqRepTest

fredoboulo:


fbacchella:
What is your IDE settings ? I'm using Eclipse, with default import sort, and results are different than yours when I launch Organise Imports. Is there any rules for that in checkstyle plugin ?

fredoboulo:
Using Intellij, the default settings as well.

PR #837, opened 2020-06-14
problem: jeromq doesn't have radio-dish implementation

somdoron:


src/main/java/zmq/socket/pubsub/Dist.java
```diff
@@ -7,7 +7,7 @@
 import zmq.Msg;
 import zmq.pipe.Pipe;
 
-class Dist
+public class Dist
```
trevorbernard, line 10: Is this meant to be public or is protected sufficient?

somdoron: I used the same modifier LB and FQ is using for consistency

PR #838, opened 2020-06-14
problem: New thread-safe sockets cannot communicate with non-thread-safe sockets

somdoron:
This is a problem, because if you want to benefit from thread safe socket in existing system you have to change the entire stack.
For system that already using a single message frame, can benefit from thread-safe socket by using the ZMQ_AS_TYPE option to make the socket use another socket type on the wire


src/main/java/zmq/io/mechanism/Mechanism.java
```diff
@@ -173,9 +175,13 @@ protected int property(String name, byte[] value)
         return 0;
     }
 
-    protected final String socketType(int socketType)
```
somdoron, line 176: It was never used, so I removed it.

PR #842, opened 2020-06-15
All thread-safe sockets

somdoron:


src/main/java/zmq/SocketBase.java
```diff
@@ -453,194 +453,221 @@ public final boolean connect(String addr)
...
-            }
+            return connectInternal(addr);
+        }
+        finally {
+            unlock();
+        }
+    }
 
-            options.mechanism.check(options);
+    public final int connectPeer(String addr)
```
somdoron, line 463: it seems like a big change, but I just changed connect to connectInternal and call it from connect and connectPeer.

fbacchella:
I think you should add new sockets type to org.zeromq.SocketType

somdoron:
You are right, this pull request only adds the new sockets to the lower-level part of the library.
I plan an additional pull request for the high-level API (which SocketType is part of). 
Some new methods are required on socket and message.

PR #894, opened 2021-08-21
Initialise SecureRandom at runtime

### LINKED ISSUE #893: Native image build failures on latest GraalVM with Random/SplittableRandom initialized at build time
I am using jeromq version `0.5.2` with [quarkus](https://quarkus.io), when I try to build a native image it throws an error:

```
Error: No instances of sun.security.provider.NativePRNG are allowed in the image heap as this class should be initialized at image runtime. To see how this object got instantiated use --trace-object-instantiation=sun.security.provider.NativePRNG.
Detailed message:
Trace: Object was reached by
        reading field java.security.SecureRandom.secureRandomSpi of
                constant java.security.SecureRandom@18595b91 reached by
        scanning method zmq.util.Utils.randomInt(Utils.java:27)
Call path from entry point to zmq.util.Utils.randomInt():
        at zmq.util.Utils.randomInt(Utils.java:27)
        at zmq.io.net.tcp.TcpConnecter.getNewReconnectIvl(TcpConnecter.java:227)
        at zmq.io.net.tcp.TcpConnecter.addReconnectTimer(TcpConnecter.java:203)
        at zmq.io.net.tcp.TcpConnecter.connectEvent(TcpConnecter.java:122)
        at zmq.poll.Poller.run(Poller.java:267)
        at java.lang.Thread.run(Thread.java:829)
        at com.oracle.svm.core.thread.JavaThreads.threadStartRoutine(JavaThreads.java:519)
        at com.oracle.svm.core.posix.thread.PosixJavaThreads.pthreadStartRoutine(PosixJavaThreads.java:192)
        at com.oracle.svm.core.code.IsolateEnterStub.PosixJavaThreads_pthreadStartRoutine_e1f4a8c0039f8337338252cd8734f63a79b5e3df(generated:0)
```
In the upcoming release of GraalVM (21.1) instances of Random/SplittableRandom class won't be allowed in the native image heap. The corresponding patch can be seen [here](https://github.com/oracle/graal/commit/11b9817fb341f9447c300304fe461933bf79c2aa#diff-8ea159555f76609d835155ca3b7f534b8a5adbb932fa6366c4db09a5b2ec2275R62).

It would be great if jeromq can initialize `java.security.SecureRandom` at runtime rather than at build time.

sriharsha-github:
Initialise SecureRandom at runtime to enable using jeromq with GraalVM. Resolves #893

### DIFF
--- a/src/main/java/zmq/util/Utils.java
+++ b/src/main/java/zmq/util/Utils.java
@@ -20,14 +20,14 @@ public class Utils
     {
     }
 
-    private static SecureRandom _random;
+    private static SecureRandom secureRandom;
     
     private static random()
     {
-        if (null == _random) {
-            _random = new SecureRandom();
+        if (null == secureRandom) {
+            secureRandom = new SecureRandom();
         }
-        return _random;
+        return secureRandom;
     }
 
     public static int randomInt()
--- a/src/main/java/zmq/util/Utils.java
+++ b/src/main/java/zmq/util/Utils.java
@@ -20,12 +20,14 @@ public class Utils
     {
     }
 
-    private static SecureRandom secureRandom;
+    private static volatile SecureRandom secureRandom;
     
     private static random()
     {
         if (null == secureRandom) {
-            secureRandom = new SecureRandom();
+            synchronized (Utils.class) {
+                secureRandom = new SecureRandom();
+            }
         }
         return secureRandom;
     }
--- a/src/main/java/zmq/util/Utils.java
+++ b/src/main/java/zmq/util/Utils.java
@@ -20,17 +20,25 @@ public class Utils
     {
     }
 
-    private static final SecureRandom random = new SecureRandom();
+    private static SecureRandom _random;
+    
+    private static random()
+    {
+        if (null == _random) {
+            _random = new SecureRandom();
+        }
+        return _random;
+    }
 
     public static int randomInt()
     {
-        return random.nextInt();
+        return random().nextInt();
     }
 
     public static byte[] randomBytes(int length)
     {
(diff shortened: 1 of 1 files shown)


fbacchella:
_ for private varialbe is not a ”java thing”, it's comming from weakly typed language. It make no sens on a strictly typed langage like java.

trevorbernard:
This code isn't threadsafe, `secureRandom` should be volatile and you would need to wrap the lazy initialization in a `synchronized` block

sjohnr:
Does anyone know if the [initialization-on-demand holder idiom](https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom) works with GraalVM? That seems like it would be the preferred idiomatic way to do this.

```java
private static class SecureRandomHolder {
    private static final SecureRandom INSTANCE = new SecureRandom();
}

private static SecureRandom random() {
    return SecureRandomHolder.INSTANCE;
}
```

If not, I think [double-checked locking](https://en.wikipedia.org/wiki/Double-checked_locking) is required (or at least preferred).

```java
private static volatile SecureRandom secureRandom;

private static SecureRandom random() {
    if (secureRandom == null) {
        synchronized (Utils.class) {
            if (secureRandom == null) {
                secureRandom = new SecureRandom();
            }
        }
    }
}
```

fredoboulo:
[ManagedContext](https://github.com/zeromq/jeromq/blob/master/src/main/java/org/zeromq/ManagedContext.java#L98) contains this idiom, I assume that someone with GraalVM could test it.
Even better, add GraalVM in the tests :)

Issue #721, opened 2019-04-18
ZMQ.Socket.close() not working all the time


Christewart:
Hi all, I'm experiencing behavior of `ZMQ.Socket.close()` not working. It is my understanding that proper tear down is closing our socket, and then calling `context.term()`

This might be related to #448 

Here is a link to the code I am running to shut down my zmq subscriber properly

https://github.com/Christewart/bitcoin-s-core/blob/361f44ce670263e7dd238ed5756bb6ea06586066/zmq/src/main/scala/org/bitcoins/zmq/ZMQSubscriber.scala#L92-L106

```scala
  def stop(): Unit = {
    //i think this could technically not work, because currently we are blocking
    //on Zmsg.recvMsg in our while loop. If we don't get another message we won't
    //be able toe evaluate the while loop again. Moving forward with this for now.
    running = false
    logger.warn(s"Attempting to close subscriber")

    subscriber.close()
    
    logger.warn(s"Terminating zmq context")
    context.term()

    logger.warn(s"Done with closing zmq")
    ()
}
```

logs:

```
2019-04-18 08:46:36.850 DEBUG o.b.d.w.CreateWallet.$anonfun$res_53$1(148) - Stopping zmq
2019-04-18 08:46:36.851 WARN  o.b.z.ZMQSubscriber.stop(97) - Attempting to close subscriber
2019-04-18 08:46:36.851 WARN  o.b.z.ZMQSubscriber.stop(101) - Terminating zmq context
```

As you can see, `context.term()` hangs forever -- which from what I understand is expected behavior if you are not closing the socket. As you can see, I am closing the socket with `subscriber.close()` so I don't know what is wrong here?

Am I missing something? I am using jermomq `0.5.1` and 
```
$ javac -version
  javac 1.8.0_191
```

daveyarwood:
The context should handle closing sockets for you, such that you only need to close the context. I believe in your code, when you call `context.term()`, it's trying to re-close all of its sockets, at least one of which (your subscriber) you have already closed manually, so maybe that's causing unexpected behavior.

Christewart:
I tried removing `subscriber.close()` and I still got the same behavior

daveyarwood:
OK, that's a good data point.

[...]

PR #919, opened 2022-08-26
fix: issue 721 deadlock bug

DayJun:
**what is the problem**

Hi all. I am using jeromq in a multithreaded environment. but I have a problem same as #721.

I have two threads in my system. One for `sub` recivier, another for `pub` sender. In the recv thread, I directly called `socket.recv()` to recv messages, which will block the thread until a message is received.

When I tried to call `context.close()`, I expect this function call will interrupt recv function, so that I can stop the recv thread directly. In most cases, it worked well. However, there are still hangs.

**how it happened**

I will use the following code to illustrate the problem.

```java
import org.zeromq.SocketType;
import org.zeromq.ZContext;
import org.zeromq.ZMQ;
import zmq.ZError;

public class TestA {
    public static void main(String[] args) throws InterruptedException {
        while(true) {
            Receiver recv = new Receiver();
            recv.run();
            Thread.sleep(100);
            recv.term();
            recv.join();
            break;
        }
    }

    static class Receiver {
        private Thread t;
        private ZContext context = new ZContext();
        public void run() {
            t = new Thread(new Runnable() {
                public void run() {     // run a thread to do recv operation
                    System.out.println("recv start.");
                    ZMQ.Socket socket = context.createSocket(SocketType.SUB);
                    socket.setLinger(0);
                    socket.connect("tcp://127.0.0.1:7899");
                    try {
                        socket.recv();
                    } catch (ZMQException e) {
                        if(e.getErrorCode() != ZError.ETERM) {
                            e.printStackTrace();
                        }
                    }
                    System.out.println("recv end.");
                }
            }, "Receiver");
            t.start();
        }

        public void term() {
            System.out.println("term start.");
            context.close();        // close the context
            System.out.println("term end.");
        }

        public void join() throws InterruptedException {
            t.join();
        }
    }
}
```

Using jeromq-0.5.3-SNAPSHOT which is compiled from master branch(`commit 2a3645dbe6161774faa5011c132907c1be3c768`).

There is a `sub` socket, and it calls recv in a single thread. Call `context.close()` in another thread to interrupt `socket.recv()`. `sub` socket's `STOP` command and this socket's `owned` socket's `TERM_ACK` command will be processed by reaper or this `sub` socket's recv function. But, if `sub` thread received `TERM_ACK` command, and then `socket.termAcks` equals to 0, `sub` socket will call `processDestroy` function which will set `sub` socket's `destroyed` field from `false` to `true`. 

But there is no `checkDestroy` function call with in `sub` socket's recv function. So at this point, reaper is waiting for `REAPED` command to remove `sub` socket from reaper, and `this.termMailbox.recv(-1L)` which is called by `context.close()` is waiting for `DONE` message which is sent by reaper's `processReaped` function. Clearly there is a deadlock here.

So this is the reason why `context.close()` hangs occasionally, and only hangs with recv socket.

**how to fix it**

```diff
diff --git a/src/main/java/zmq/SocketBase.java b/src/main/java/zmq/SocketBase.java
index 2de11bf..49f8548 100644
--- a/src/main/java/zmq/SocketBase.java
+++ b/src/main/java/zmq/SocketBase.java
@@ -1135,10 +1135,16 @@ public abstract class SocketBase extends Own implements IPollEvents, Pipe.IPipeE

         //  Initialize the termination and check whether it can be deallocated
         //  immediately.
+        setCtxTerminated();
         terminate();
         checkDestroy();
     }

+       private final void setCtxTerminated()
+    {
+        ctxTerminated = true;
+    }
+
     //  Processes commands sent to this socket (if any). If timeout is -1,
     //  returns only after at least one command was processed.
     //  If throttle argument is true, commands are processed at most once
@@ -1212,7 +1218,7 @@ public abstract class SocketBase extends Own implements IPollEvents, Pipe.IPipeE
         try {
             monitorSync.lock();
             stopMonitor();
-            ctxTerminated = true;
+            setCtxTerminated();
         }
         finally {
             monitorSync.unlock();
```

I think the root cause of the hang is, when a socket send `REAP` message to reaper, the reaper should takes over the processing of commands of this socket completely. But after called `processReap -> startReaping`, the socket can still recving, and currently progressed recv operation will not be interrupted. If socket is still recving, it may receive commands that should have been handled by reaper (such as the last `TERM_ACK`). 

Reaper will call `inEvent` to process commands:

```java
@Override
    public final void inEvent()
    {
        //  This function is invoked only once the socket is running in the context
        //  of the reaper thread. Process any commands from other threads/sockets
        //  that may be available at the moment. Ultimately, the socket will
        //  be destroyed.

        lock();

        try {
            //  If the socket is thread safe we need to unsignal the reaper signaler
            if (threadSafe) {
                reaperSignaler.recv();
            }

            processCommands(0, false, null);

        }
        finally {
            unlock();
        }

        checkDestroy();
    }
```

Common sockets will directly call `processCommands` in `recv` function:

```java
public final Msg recv(int flags, AtomicBoolean canceled)
    {
        lock();

        try {
            //  Check whether the library haven't been shut down yet.
            if (ctxTerminated) {
                errno.set(ZError.ETERM);
                return null;
            }

            ......
            while (true) {
                if (!processCommands(block ? timeout : 0, false, canceled)) {
                    return null;
                }
                ......
            }

            extractFlags(msg);
            return msg;
        }
        finally {
            unlock();
        }
    }
```

So I just force common socket don't recv messages when reaping, and this will solve the problem.

### DIFF
--- a/src/main/java/zmq/SocketBase.java
+++ b/src/main/java/zmq/SocketBase.java
@@ -1140,7 +1140,7 @@ public abstract class SocketBase extends Own implements IPollEvents, Pipe.IPipeE
         checkDestroy();
     }
 
-	private final void setCtxTerminated()
+    private final void setCtxTerminated()
     {
         ctxTerminated.set(true);
     }
--- a/src/main/java/zmq/SocketBase.java
+++ b/src/main/java/zmq/SocketBase.java
@@ -1,15 +1,5 @@
 package zmq;
 
-import java.io.IOException;
-import java.net.InetSocketAddress;
-import java.nio.channels.SelectableChannel;
-import java.nio.channels.SocketChannel;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.Set;
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.locks.ReentrantLock;
-
 import zmq.io.IOThread;
 import zmq.io.SessionBase;
 import zmq.io.net.Address;
@@ -23,6 +13,16 @@ import zmq.socket.Sockets;
 import zmq.util.Blob;
 import zmq.util.Clock;
 import zmq.util.MultiMap;
+
+import java.io.IOException;
+import java.net.InetSocketAddress;
+import java.nio.channels.SelectableChannel;
+import java.nio.channels.SocketChannel;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.locks.ReentrantLock;
 
 public abstract class SocketBase extends Own implements IPollEvents, Pipe.IPipeEvents
 {
@@ -57,10 +57,13 @@ public abstract class SocketBase extends Own implements IPollEvents, Pipe.IPipeE
     //  If true, associated context was already terminated.
     private AtomicBoolean ctxTerminated;
 
+    // If processCommand function was called from InEvent function.
+    private ThreadLocal<Boolean> isInEventThreadLocal;
+
     //  If true, object should have been already destroyed. However,
     //  destruction is delayed while we unwind the stack to the point
     //  where it doesn't intersect the object being destroyed.
-    private boolean destroyed;
+    private AtomicBoolean destroyed;
 
     //  Socket's mailbox object.
     private final IMailbox mailbox;
@@ -114,7 +117,8 @@ public abstract class SocketBase extends Own implements IPollEvents, Pipe.IPipeE
         super(parent, tid);
         active = true;
         ctxTerminated = new AtomicBoolean();
-        destroyed = false;
+        isInEventThreadLocal = new ThreadLocal<>();
+        destroyed = new AtomicBoolean();
         lastTsc = 0;
         ticks = 0;
         rcvmore = false;
@@ -175,7 +179,7 @@ public abstract class SocketBase extends Own implements IPollEvents, Pipe.IPipeE
             }
 
             stopMonitor();
-            assert (destroyed);
+            assert (destroyed.get());
         }
         finally {
             monitorSync.unlock();
@@ -1135,14 +1139,14 @@ public abstract class SocketBase extends Own implements IPollEvents, Pipe.IPipeE
 
         //  Initialize the termination and check whether it can be deallocated
         //  immediately.
-        setCtxTerminated();
+        ctxTerminated.set(true);
         terminate();
         checkDestroy();
     }
 
-    private final void setCtxTerminated()
-    {
-        ctxTerminated.set(true);
+    private boolean isInEvent() {
+        Boolean bRes = isInEventThreadLocal.get();
+        return null != bRes && bRes;
     }
 
     //  Processes commands sent to this socket (if any). If timeout is -1,
@@ -1189,6 +1193,10 @@ public abstract class SocketBase extends Own implements IPollEvents, Pipe.IPipeE
             cmd = mailbox.recv(0);
         }
 
+        if(!isInEvent() && destroyed.get()) {
+            sendReapAck();
+        }
+
         if (errno.get() == ZError.EINTR) {
             return false;
         }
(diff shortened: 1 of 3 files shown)


src/main/java/zmq/SocketBase.java
```diff
@@ -1135,10 +1135,16 @@ final void startReaping(Poller poller)
...
+        setCtxTerminated();
         terminate();
         checkDestroy();
     }
 
+	private final void setCtxTerminated()
+    {
+        ctxTerminated = true;
+    }
+
```
trevorbernard, line 1147: I don't think this is sufficient in solving the deadlock. It's still not atomically setting the boolean value. Consider making ctxTerminated a volatile or changing it to be AtomicBoolean. I'd prefer the latter it's easier to make correct.

DayJun: You are right, and new codes were committed just now. Thanks for your reply.

DayJun:
Hold on, the bug is not fixed completely, I still didn't interrupt the recv operation, whitch will block at 

```java
Mailbox.java

@Override
    public Command recv(long timeout)
    {
     ......

        //  Wait for signal from the command sender.
        boolean rc = signaler.waitEvent(timeout);
        ......
        return cmd;
    }
```

DayJun:
I added a new Command called `REAP_ACK`, It's sent by non-reaper socket to reaper to check destroy.

DayJun:
> Hold on, the bug is not fixed completely, I still didn't interrupt the recv operation, which will block at
> 
> ```java
> Mailbox.java
> 
> @Override
>     public Command recv(long timeout)
>     {
>      ......
> 
>         //  Wait for signal from the command sender.
>         boolean rc = signaler.waitEvent(timeout);
>         ......
>         return cmd;
>     }
> ```

nzbnizebin:
When I upgraded the version of jeromq from 0.5.2 to 0.5.3, the deadlock issue still occurs in my scenario. I believe the root cause is that a series of stop commands are handled by recv method exceptionally instead of the inEvent method. In multithreaded scenarios, it should be ensured that the thread blocking on recv method exits first (and thanks for interrupt work) before closing the ZContext  . I hope this can be documented as a sample.
```java
Breakpoint reached
	at zmq.SocketBase.processDestroy(SocketBase.java:1006)
	at zmq.Own.checkTermAcks(Own.java:254)
	at zmq.Own.unregisterTermAck(Own.java:232)
	at zmq.SocketBase.pipeTerminated(SocketBase.java:1126)
	at zmq.pipe.Pipe.processPipeTermAck(Pipe.java:397)
	at zmq.ZObject.processCommand(ZObject.java:91)
	at zmq.Command.process(Command.java:75)
	at zmq.SocketBase.processCommands(SocketBase.java:942)    // exceptionally processCommands
	at zmq.SocketBase.recv(SocketBase.java:834)
	at zmq.ZMQ$Event.read(ZMQ.java:253)
	at org.zeromq.ZMQ$Event.recv(ZMQ.java:4055)
	at …..
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.lang.Thread.run(Thread.java:833)
Breakpoint reached
	at zmq.SocketBase.processDestroy(SocketBase.java:1006)
	at zmq.Own.checkTermAcks(Own.java:254)
	at zmq.Own.unregisterTermAck(Own.java:232)
	at zmq.SocketBase.pipeTerminated(SocketBase.java:1126)
	at zmq.pipe.Pipe.processPipeTermAck(Pipe.java:397)
	at zmq.ZObject.processCommand(ZObject.java:91)
	at zmq.Command.process(Command.java:75)
	at zmq.SocketBase.processCommands(SocketBase.java:942)   // expected  processCommands
	at zmq.SocketBase.inEvent(SocketBase.java:1065)
	at zmq.poll.Poller.run(Poller.java:273)
	at java.lang.Thread.run(Thread.java:833)

```

Issue #905, opened 2021-12-03
Exception while processing command leave IOThread in bad state (Not invoking inEvent to serve new command)


WeiliangLuo:
It is not obvious, but I recently find out that **the IOThread has to drain all commands in its mailbox in order to be signaled again for new commands**. This is because draining the messages off underlying YPipe [will reset the YPipe.c to -1](https://github.com/zeromq/jeromq/blob/bda7e6f997cafae7abcf734b2f8da6285a5eac05/src/main/java/zmq/pipe/YPipe.java#L111), which cause ["flush()" to return false ](https://github.com/zeromq/jeromq/blob/bda7e6f997cafae7abcf734b2f8da6285a5eac05/src/main/java/zmq/pipe/YPipe.java#L80)on sender, and only then [the sender of new command will signal the IOThread](https://github.com/zeromq/jeromq/blob/bda7e6f997cafae7abcf734b2f8da6285a5eac05/src/main/java/zmq/Mailbox.java#L70) to invoke inEvent().

However, an exception while [IOThread is processing the command](https://github.com/zeromq/jeromq/blob/bda7e6f997cafae7abcf734b2f8da6285a5eac05/src/main/java/zmq/io/IOThread.java#L80) will stop it from draining its mailbox. As a result, the IOThread will be left in running state but never waked up again to serve inEvent even though there are new commands sent to it.

The exception we hit is the same as the one mentioned in https://github.com/zeromq/jeromq/issues/871.
In our case, we understand that it is caused by a race condition between worker thread notifying the IOThread to ActivateWrite on a backed up socket and the heartbeat reconnecting to the remote end. This results the ActivateWrite command arrives to a new StreamEngine object just set up for the new connection, which is not fully initialized pending handshake.
We learnt that it is not a good idea to enable heartbeat on Pull socket as the back pressure will trigger the heartbeat to reconnect frequently (given a reasonably low heartbeat timeout).

I see few different ways to handle the exception better, for example, let the IOThread die instead of fail silently, handle the exception inside the inEvent or even call iothread.inEvent without the signal. 
I am curious on your thoughts on this.

Thanks!

fbacchella:
Do you think this commit might help you ? https://github.com/zeromq/jeromq/commit/ece14303872354e689065ff635086ca355336d47

fbacchella:
One more question. Are you able to reproduce it ? A NPE is always a bug. So it should be resolver too.

ben-ng:
[...]

PR #939, opened 2023-01-16
Improved concurrency and thread-safeness in Mailbox

fbacchella:
Simplified Mailbox, always thread safe, so MailboxSafe usages are removed and the class is marked at deprecated.

### DIFF
--- a/src/main/java/zmq/Mailbox.java
+++ b/src/main/java/zmq/Mailbox.java
@@ -2,29 +2,19 @@ package zmq;
 
 import java.io.IOException;
 import java.nio.channels.SelectableChannel;
-import java.util.concurrent.locks.Lock;
-import java.util.concurrent.locks.ReentrantLock;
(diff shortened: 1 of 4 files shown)


trevorbernard:
Since Sockets are known to be to be not thread safe, what is the purpose if making internals thread safe? It might the case we're adding extra work for no material gain.

trevorbernard:
This PR satisfies all c4 requirements so I must merge but it's worth considering my question. I do not know the answer.

fbacchella:
Mailbox was using YPipe, a class that is really complicated to understand and ensure to be correct. It this class could be totally removed, it would be really better for reliability of jeromq. I’m also investigating #905 and I’m chasing down many occurrence of non compliance of java memory model, where a non-protected or non volatile variable could be access with wrong values.
MailboxSafe was handling concurrency in a custom way, using a standard class should improve reliability too.

PR #952, opened 2023-04-07
Code refactoring implemented to remove smells

piyush-akoliya:
Implemented following code refactoring techniques:
Extract method
Decompose conditional
Introduce explaining variable

Move method
Pull-up variable
Extract class


fbacchella:
Did you run a `mvn compile -Pcheckstyle` on your code ? I don’t think so.

trevorbernard:
This PR is doing way too much. I'm still unclear what problems it aims to solve. As @fbacchella we already mentioned we have a coding style that jeromq uses.

PR #966, opened 2023-11-28
Refactored few implementation and design code smells. 

KrishnaVaibhavY:
### **Implementation Smells**
**Extract method**
Extracted ‘setupBroadcastChannel’ and ‘broadcastBeacon’ in BroadcastClient, and for better code organization and readability. This resolves the one of the implementation smell in the ‘ZBeacon’ class 

**Rename method/variable**
Renamed handle to datagramChannel and isRunning to running for better clarity.

**Decompose conditional**
Decomposed the complex conditional in BroadcastServer's run method into shouldProcessMessage. Also introduced beaconData in BroadcastClient's broadcastBeacon for clarity.

### **Design Smells** 

**Pull-up variable/method**
Moved the code field and methods related to it (getErrorCode() and a part of toString()) from ZMQException to UncheckedZMQException. This allows any subclass of UncheckedZMQException to have an error code and the associated methods.

**Push-down variable/method**

The SimpleCurveAuth inner class has a dependency on ZCertStore.Fingerprinter. We can push down the fingerprinter initialization from the ZAuth constructor to SimpleCurveAuth. This makes SimpleCurveAuth more independent and encapsulates its initialization.
From ZAuth to SimplePlainAuth:

SimplePlainAuth uses properties for passwords. These properties are specific to SimplePlainAuth and can be initialized inside it, rather than being passed from ZAuth.

**Replace conditional with polymorphism**

ZAuth uses a map auths to handle different authentication mechanisms. This can be improved by using polymorphism. Each Auth implementation (SimpleCurveAuth, SimplePlainAuth, SimpleNullAuth) can override a method like handleAuth(ZapRequest request) to encapsulate the authentication logic.

### DIFF
--- a/src/main/java/org/zeromq/UncheckedZMQException.java
+++ b/src/main/java/org/zeromq/UncheckedZMQException.java
@@ -1,26 +1,42 @@
 package org.zeromq;
 
-public abstract class UncheckedZMQException extends RuntimeException
-{
+public abstract class UncheckedZMQException extends RuntimeException {
     private static final long serialVersionUID = 1L;
 
-    public UncheckedZMQException()
-    {
+    protected int code; // Pulled-up variable
+
+    public UncheckedZMQException() {
         super();
     }
 
-    public UncheckedZMQException(String message)
-    {
+    public UncheckedZMQException(String message) {
         super(message);
     }
 
-    public UncheckedZMQException(Throwable cause)
-    {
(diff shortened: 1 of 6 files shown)


trevorbernard:
Also, as per the C4 process, a patch should be a minimal and accurate answer to exactly one identified and agreed problem. This PR implements many fixes to many problems. Due to it's staleness, I'll be closing this issue.

PR #969, opened 2024-01-29
fix typo in doc/sockets.md

deschcla:
add missing links for zguide


fbacchella:
I don’t get the purpose of this modification, it adds a link to the generic `ØMQ - The Guide`. Why this link in particular ? The purpose of the modified document is using ZMQ sockets in general, not be be a generic documentation about ZMQ. Also the link was added as http instead of https, that is a deprecated practice. And worse, it’s broken any way, so it’s useless.

deschcla:
Hello @fbacchella ! 

> Why this particular link? 

I've just added the missing link to the site referenced by the "See also" section, and I've added the same link as in the documentation `pollers.md` since the reference is the same, which explains the `http`. However, if this link doesn't seem useful to you in this documentation, it might be preferable to remove it completely.

fbacchella:
The link to poller should stay in the poller documentation. But indeed the link is broken there too, so that would be a good idea to fix it there.

PR #1004, opened 2024-10-25
修复系统时间跳变引起的长时间无返回

zhanzhihongzzh:
系统时间突然跳到过去某个时间，会导致这个 timeout 非常大，引起长时间没有返回
timeout = (int) (end - Clock.nowMS())

### DIFF
--- a/jeromq-core/src/main/java/zmq/SocketBase.java
+++ b/jeromq-core/src/main/java/zmq/SocketBase.java
@@ -834,7 +834,7 @@ public abstract class SocketBase extends Own implements IPollEvents, Pipe.IPipeE
             //  Compute the time when the timeout should occur.
             //  If the timeout is infinite, don't care.
             int timeout = options.sendTimeout;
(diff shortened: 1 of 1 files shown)


fbacchella:
In english please, what are you talking about ?

fbacchella:
`Clock.nowMS()` uses `System.currentTimeMillis()`, which uses one of the internal OS clock.
`System.nanoTime()` uses `System.nanoTime()`, which is an internal mesure of the VM.

If you set the time of the OS (which should never be done) the first one fails to mesure elapsed time. And so the timeout fails. But any way, i don’t think the patch is good, it uses a division, that is a heavy operation that should be avoided. Is it the only place where a timeout is handled in such a way in the all library ? That’s a quick fix solution, not a comprehensive one.

trevorbernard:
We shouldn't pass judgement on a valid PR according to the C4, but this isn't a valid PR because it doesn't solve an agreed upon problem.

